PR #23, opened 2015-09-02
fix the test situation.

reason-bv:
* The tests were always exiting with error code 0.
* The tests reporting was exiting without waiting for completion, much more noticeable once async tests were added.

The problem was probably karma-tap, so this PR fixes the problems above by moving to Mocha/Chai.


tnunamak:
These changes look sound and the tests pass for me. I'm not very familiar with karma or tap, so I don't really have an opinion about the decision to switch to mocha/chai instead. Aside from that, :+1: for me.

PR #24, opened 2015-09-02
Pull over the easy four modules from scoutfile.

reason-bv:
This PR pulls in the scoutfile modules that should come over basically unchanged.

* cookie
* evented
* ie
* loader

The work is all in rewriting the tests, so that's where the attention should go.

### DIFF
--- a/lib/loader/index.js
+++ b/lib/loader/index.js
@@ -1,0 +1,244 @@
+/* eslint no-use-before-define: 0 */
+/**
+ * @fileOverview
+ * Load JavaScript and CSS at runtime.
+ *
+ * This module is derived from yepnope.js 2.0.0:
+ * Copyright (c) 2014, Alex Sexton
+ * Copyright (c) 2014, Ralph Holzmann
+ * All rights reserved.
+ * New BSD license:
+ * https://github.com/SlexAxton/yepnope.js/blob/master/LICENSE.md
+ *
+ * @module
+ */
+
+var global = require('../global');
+var doc = global.document;
+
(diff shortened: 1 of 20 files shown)


test/unit/loader/index.spec.js
```diff
@@ -0,0 +1,395 @@
...
+
+      it('callback must be a function', function () {
+        expect(function () {
+          loader.loadStyleSheet('/base/test/fixtures/lib.loader.loadstylesheet.css', {}, 123);
+        }).to.throw(/`callback` must be a function/);
+      });
+
+      /*
+      // TODO: Fix this test; it fails, even though the actual functionality
+      // appears to work fine. Tracked as #20.
```
tnunamak: I'm guessing this comment isn't valid anymore?

lib/loader/index.js
```diff
@@ -0,0 +1,254 @@
...
+      clearTimeout(timeoutHandle);
+      script.onload = script.onreadystatechange = script.onerror = null;
+      script.parentNode.removeChild(script);
+
+      if (callback) {
+        try {
+          callback(err);
+        }
+        catch (e) {
+          // FIXME: report this error once analytics support is available
```
tnunamak: Maybe not applicable to this PR specifically, but surely there's something we can do to avoid swallowing these exceptions completely? I'm happy to open a PR if someone has an idea about how to do that (logging, wrapping the error, etc.).

reason-bv:
@tnunamak Updated code in response to comments.

tnunamak:
:+1: for me

PR #26, opened 2015-09-03
Add namespacer module.

reason-bv:
A pull request to add a namespacer module.

PR #27, opened 2015-09-04
Add application module.

reason-bv:
Pull over the application module.

This PR is based on top of https://github.com/bazaarvoice/bv-ui-core/pull/26 because the documentation talks about both application and namespacer modules, even though the code is independent.

### DIFF
--- a/lib/application/README.md
+++ b/lib/application/README.md
@@ -1,0 +1,64 @@
+# application
+
+The application module provides an Application class that provides methods to
+queue your users' calls until the actual code to support those calls is
+present.
+
+For example, your user may expect to be able to ask your application to render
+immediately after this initial application module is loaded, even though the
+code to perform the rendering has not yet loaded.
+
+The application module is typically used in conjunction with the
+[loader][1] module, [namespacer][2] module, and some configuration data, as
+shown here. Once you create an application instance, it is good practice to add
+it as a property to a namespace:
+
+```js
+var Application = require('bv-ui-core/lib/application');
+var loader = require('bv-ui-core/lib/loader');
+var namespacer = require('bv-ui-core/lib/loader');
+var config = require('json!config.json');
+
+// Creates a namespace at window.APP and an application object at
+// window.APP.exampleApp. The config object will be available at
+// window.APP.exampleApp.config.
+namespacer.namespace('APP').registerProperty(
+  'exampleApp',
+  new Application(config)
+);
+
+// Now load following code.
+// loader.load(url).then( ... do something ... );
+```
+
+## Methods to be Called by Users
+
+An application instance provides the following methods:
+
+- `render(config)`: A method intended for your users to call any time after
+  this application module is loaded, passing in configuration information
+  related to a request to render your application.
+
+```js
+APP.exampleApp.render({ productId : 1 });
+```
+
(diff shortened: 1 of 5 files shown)


lib/application/index.js
```diff
@@ -0,0 +1,113 @@
...
+ */
+
+// --------------------------------------------------------------------------
+// Functions.
+// --------------------------------------------------------------------------
+
+/**
+ * Process a queue.
+ *
+ * @param {Mixed[]} queue An array of queued items.
```
taylormck: Super tiny nitpick (seriously, on a scale from 0 to 10, this is a 0.00001): if we'd like to keep with GCC's annotation, `Mixed[]` should be replaced with `Array`, according to [Google's type expressions](https://developers.google.com/closure/compiler/docs/js-for-compiler?hl=en#types).
Then again, if we're not at all concerned with GCC annotation, then I'm not at all worried about it.

lib/application/README.md
```diff
@@ -0,0 +1,64 @@
...
+The application module provides an Application class that provides methods to
+queue your users' calls until the actual code to support those calls is
+present.
+
+For example, your user may expect to be able to ask your application to render
+immediately after your scout file is loaded, even though the code to perform
+the rendering has not yet loaded.
+
+Once you create an application instance, it is available via its application
+name as a property in the scout file's namespace. (By default, the scout
```
tnunamak: I think it's worth linking to the scoutfile repo somewhere in here if "scout file" in the bv-ui-core docs specifically means that library.

lib/application/README.md
```diff
@@ -0,0 +1,64 @@
...
+// Creates a namespace at window.APP and an application object at
+// window.APP.exampleApp. The config object will be available at
+// window.APP.exampleApp.config.
+namespacer.namespace('APP').registerProperty(
+  'exampleApp',
+  new Application(config)
+);
+
+// Load the core application, which will define an APP.exampleApp.onLoad method.
+loader.loadScript(config.appJS).then(APP.exampleApp.onLoad);
```
tnunamak: Would `onLoad` be an Application API? I don't see it anywhere.

lib/application/README.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+An application instance provides the following methods:
```
tnunamak: This is just a nitpick: I think these docs would be much clearer if the API description was split into "users call these" and "your application calls these" sections.

lib/application/index.js
reason-bv: I wasn't aware we had standardized to any one set of type expressions. Have we?

reason-bv:
@tnunamak Addressed your comments in the application README.md.

tnunamak:
Thanks, looks great :+1:.

lib/application/index.js
taylormck: We haven't afaik, but if we want to, we should decide now.
Otherwise, this is a non-issue.

I'm not saying we should, but I figured it worth mentioning.

rmurphey: If we want to attempt to standardize on this, please open an issue or pull request on the web-apps-team repo explaining what we want to standardize on and the rationale for doing so.

PR #28, opened 2015-09-10
Migrating bv-ui-pixels-displayed to module in the core

msmolev:
This is an attempt to migrate https://github.com/bazaarvoice/bv-ui-pixels-displayed into the core library.
Original library, however, only checked vertical visibility. So I have also duplicated the function and created horizontal version as well. 

Migrated tests, created horizontal tests, added link in Readme.

### DIFF
--- a/lib/pixelsDisplayed/README.md
+++ b/lib/pixelsDisplayed/README.md
@@ -1,0 +1,19 @@
+# pixelsDisplayed
+
+> A module for detecting whether an element is located within the viewport
+
+## Usage
+
+```js
+var bvUIPixelsDisplayed = require('bv-ui-core/lib/pixelsDisplayed');
+var pixelsDisplayedVertical = bvUIPixelsDisplayed.pixelsDisplayedVertical;
+var pixelsDisplayedHorizontal = bvUIPixelsDisplayed.pixelsDisplayedHorizontal;
+
+
+if (pixelsDisplayedVertical(element) > 0) {
+  // element is vertically within viewport
+  if (pixelsDisplayedHorizontal(element) > 0) {
+    // element is both vertically and horizontally within viewport
+  }
+}
+```
--- a/lib/pixelsDisplayed/index.js
+++ b/lib/pixelsDisplayed/index.js
@@ -1,0 +1,113 @@
+/**
+ * @fileOverview Checks visibility of an element by calculating the
+ * visible pixels for vertical and horizontal viewport.
+ * This is a migration of repository bv-ui-pixels-displayed
+ */
+var global = require('../global');
+
+function getDimensions () {
+  var window = global;
+  var docEl = window.document && window.document.documentElement;
+  return {
+    height: window.innerHeight || docEl.clientHeight,
+    width: window.innerWidth || docEl.clientWidth
+  };
+}
+
+function pixelsDisplayedVertical (el) {
+  var rect = el.getBoundingClientRect();
+  var displayed;
+  var windowHeight = module.exports.getDimensions().height;
+
+  // There are five potential DOM node positions relative to the viewport:
+  // * Completely above the viewport, not visible
+  // * Completely below the viewport, not visible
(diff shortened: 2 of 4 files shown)


lib/pixelsDisplayed/README.md
```diff
@@ -0,0 +1,28 @@
+# pixelsDisplayed
+
+> A module for detecting whether an element is located within the viewport
+
+## Module arguments
+
+ - `window`: Reference to window.
```
reason-bv: It won't need either of these, as it can use `global`.

lib/pixelsDisplayed/index.js
```diff
@@ -0,0 +1,107 @@
+/**
+ * @fileOverview Checks visibility of an element by calculating the
+ * visible pixels for vertical and horizontal viewport.
+ * This is a migration of repository bv-ui-pixels-displayed
+ */
+module.exports = function (window, document) {
```
reason-bv:
```
var global = import('../global');

module.exports = function () {
  var window = global;
  var document = window.document;
```

test/unit/pixelsDisplayed/index.spec.js
```diff
@@ -0,0 +1,251 @@
+/**
```
reason-bv: Some small changes needed here to mock the the necessary items given use of global rather than arguments.

msmolev:
Updated structure to pull out the accessor for the testing purposes

lib/pixelsDisplayed/index.js
```diff
@@ -0,0 +1,113 @@
...
+  // At this point, the only remaining case is that the left of the DOM node is
+  // somewhere in the viewport. We can determine the maximum possible amount
+  // displayed rather easily, by determining how much distance there is between
+  // the left of the bounding rectangle and the viewport's width.
+  // However, if this is larger than our DOM node, we don't want to report more
+  // displayed than the node's full width;
+  return Math.min(windowWidth - rect.left, rect.width);
+}
+
+module.exports = { getDimensions: getDimensions,
```
reason-bv: If you do this then if will never update as the window updates. You should export the functions.

reason-bv: Ah, never mind, they are the functions.

test/unit/pixelsDisplayed/index.spec.js
```diff
@@ -0,0 +1,273 @@
...
+
+  var el = {
+    getBoundingClientRect: function () {
+      return rect;
+    }
+  };
+
+  function stubAccessor () {
+    sinonSandbox = sinon.sandbox.create();
+    spy = sinonSandbox.spy();
```
reason-bv: What's the intent here; the stub interface includes the spy interface, so anything you want to do with spy can be done with the stub.

test/unit/pixelsDisplayed/index.spec.js
```diff
@@ -0,0 +1,273 @@
...
+      height: 200,
+      top: 100,
+      width: 200,
+      left: 100
+    };
+  });
+
+  afterEach(function () {
+    // we do expect spy to have been called during each test
+    expect(spy.called).to.eql(true);
```
reason-bv:
```
sinon.assert.called(module.getDimensions)
```

lib/pixelsDisplayed/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+if (pixelsDisplayedVertical(element) > 0) {
+  // element is vertically within viewport
+  if (pixelsDisplayedHorizontal(element) > 0) {
+    // element is both vertically and horizontally within viewport
+  }
+}
+```
+
+## License
```
reason-bv: The license block can be deleted - the license for the whole package is in the upper directory.

reason-bv:
:+1: Aside from items noted.

msmolev:
Corrected the license notice

Issue #30, opened 2015-10-08
Logging solution


sirugh:
It seems like we don't have a recommended logging module for front-end apps.

console.log is discouraged, but it would be convenient to have a module here that wrapped that behavior safely.

rmurphey:
Things I would want from this:

- behaves a whole lot like console.log
- silent in production
- can be enabled in production by setting a cookie

sirugh:
Suggestion by @benhollander: https://github.com/pimterry/loglevel

brianarn:
Seems like loglevel supports the desired goals of the project, aside from the production enabling via cookie. We have other code around BV that supports this pattern, and at a cursory glance, it looks like loglevel would support a plugin wherein we could introduce that behavior.

That being said, we do also have a more comprehensive logging solution in another project, which already supports all of the above criteria, along with deeper support for `console` similarness. It appears that loglevel doesn't support things like `group`, `time`, or `assert`.

It does support the idea of unique loggers, whereas our internal version is a single instance for everyone, which could be interesting (though I'm not sure how much value that would bring in practice).

rmurphey:
Any further thoughts on this? What do we want logging to look like? I'd love to see a shared logging solution here.

sirugh:
loglevel sounds fine if we can write a plugin to wrap its usage.

benhollander:
>That being said, we do also have a more comprehensive logging solution in another project, 

What's that?

sirugh:
I believe he is referring to BVReporter: https://github.com/bazaarvoice/firebird/blob/master/static/dev/scripts/framework/util/bvreporter.js

benhollander:
If that's already vetted and something that we already like, I'm :+1: for using it more.

brianarn:
Given that BVReporter has been pretty vetted and meets all of @rmurphey's asks above, we could easily abstract it out for use in the repo. That was the piece of code I was referring to.

sirugh:
https://github.com/bazaarvoice/bv-ui-core/compare/bazaarvoice:master...sirugh:logger?expand=1

PR #32, opened 2015-10-23
Logger

sirugh:
A port of BVReporter for ui-core.

#30


lib/logger/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+### group/groupEnd
+https://developer.mozilla.org/en-US/docs/Web/API/console/group
+
+### time/timeEnd
+https://developer.mozilla.org/en-US/docs/Web/API/console/time
+
+### assert
+
+Assertions can be made if `logLevel` is not `Logger.OFF`. You can assert a function's result or a boolean value.
```
rmurphey: Assertions can be expensive -- we may not want to run them in production, e.g. when we're using `Logger.ERROR` as the log level. Interested what others think here as to whether assertions should be tied to the log level, and if so, what level.

lib/logger/index.js
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    if (!assertion) {
+      throw new Error('Assertion failed: ' + message);
+    }
+  }
+};
+
+function logger (level, args) {
+  if (logLevel === Logger.OFF) {
```
rmurphey: Here we use `Logger.OFF` but above we use explicit values. I think we should use `Logger` properties throughout.

lib/logger/index.js
```diff
@@ -0,0 +1,167 @@
...
+
+  if (ie) {
+    console.log(args);
+    return;
+  }
+
+  try {
+    (console[level] || console.log).apply(console, args);
+  }
+  catch (e) {}
```
tnunamak: I understand why, if logging fails, there might be very little left you can do if something goes wrong. Still, is there anything we can do here instead of swallowing the exception? Should we swallow at all? If this is really the end of the line, a comment explaining why it's okay to swallow here would be nice.

sirugh: If we can't log, there is literally nothing we can do with the code in its current state. In the future it may be beneficial to allow a configurable log consumer that we can publish logs to and in this case the error. However, that hasn't been implemented and it may add a lot of overhead to what is otherwise a pretty simple module.

lib/logger/README.md
sirugh: In BVReporter assertions are only allowed if the cookie is found or if you are on localhost (basically `isDev`). Since we don't have a concept of environment in this module, the only way we can check is via either the cookie or the level set by the user.

I am however concerned that there may be assertions made in the code BEFORE the user of this module is able to set the log level.

lib/logger/index.js
tnunamak: > a configurable log consumer that we can publish logs to and in this case the error

This is an interesting idea, it would be a good place to hook in analytics.

We don't want to error out here because IE, right? Still, can I get a comment? :smile:

lib/logger/README.md
rmurphey: This also brings up that this doesn't seem to meet one of the goals: to be able to enable this in production via cookie.

rmurphey: This also brings up *another* thought: we should probably have a separate assertion module.

lib/logger/index.js
sirugh: Done.

lib/logger/README.md
sirugh: Does this not? https://github.com/bazaarvoice/bv-ui-core/pull/32/files#diff-f8f0b21002d80d85a7256fac15628da4R32

If a page has the specified cookie present it will automatically set the level. That of course can be overwritten later by anyone calling Logger.setLogLevel(). Do we want the cookie to take precedence over whatever the code is doing?

lib/logger/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Logger.log('Hello, World!') // Prints Hello, World!
+
+Logger.setLogLevel(Logger.INFO)
+
+Logger.log('Hello, World!') // Prints nothing because of logLevel
+```
+
+### group/groupEnd
+https://developer.mozilla.org/en-US/docs/Web/API/console/group
```
tnunamak: This is a little nitpicky but just to let you know, it wasn't immediately clear to me that group was a method I could call on Logger. Then I realized it was implied that each sub-header here is a method. Also, the first one starts with a '#', the others don't.

test/unit/logger/index.spec.js
```diff
@@ -0,0 +1,106 @@
...
+    global.console = false;
+
+    Logger.info('foo');
+    expect(infoStub).to.not.have.been.called;
+
+    global.console = tempConsole;
+  });
+
+  it.skip('defaults to `log` in internet explorer', function () {
+    // TODO - fake IE
```
sirugh: I'd like to do this, but not sure how at this time.

lib/logger/index.js
```diff
@@ -0,0 +1,171 @@
...
+var ie = require('../ie')
+var cookie = require('../cookie');
+
+var logLevel = 1; // Default to INFO.
+
+/**
+ * Set the logLevel to whatever the _bvui_debug cookie has stored.
+ */
+var cookieObj = {};
+var debugCookie = cookie.read('_bvui_debug');
```
rmurphey: Sorry, I completely glanced over this in my first read.

PR #34, opened 2015-10-26
correct docs for loader module

rmurphey:
The docs for the loader module said that loader methods return a promise; they do not. They also showed that the callback to loader methods receives two arguments; it receives one. This PR addresses these issues.


brianarn:
:+1:

PR #47, opened 2015-12-06
Add logger module to readme

nason:
I came looking for the logger, but couldn't find it in the readme :fries:


sirugh:
Oops! :+1:

PR #48, opened 2015-12-06
Logger group and groupEnd arguments slice fix

nason:
Calls to `Logger.group` and `Logger.groupEnd` were throwing errors. I noticed there weren't tests for these, so I added them too.


sirugh:
:+1: Thanks for finding and fixing this @nason!

PR #51, opened 2016-01-20
JST-7: Add region module

taylormck:
## What's in this PR ?

New laws in the EU require that identifying information stay physically located in the EU. In order to comply with this, we'll need to be able to tell when we're running in the EU. It has been determined that the onus of providing the correct locale is on the client, so we can safely rely on the base locale of an implementation.

This PR adds a module that UI projects can use to determine if it's in EU or not. It has been designed with extensibility in mind, in case we want to add support for detecting other regions in the future.

## How to verify

1. Check out this PR
1. In the `bv-ui-core` repo, run `node`.
1. Run `var region = require('./lib/region');` to get a reference to the root module.
1. Run `region.hasLocale('europe', 'en_US');`. Should return `false`.
1. Run `region.hasTerritory('europe', 'US');`. Should return `false`.
1. Run `region.hasLocale('europe', 'en_GB');`. Should return `true`.
1. Run `region.hasTerritory('europe', 'GB');`. Should return `true`.
1. Run `region.lookupLocale('en_GB'); Should return `'europe'`.
1. Run `region.lookupLocale('en_US'); Should return `'unsupported'`.
1. If you prefer, instead of the "generic" module, you can run `var europe = require('./lib/region/europe');` to load the European specific module. Useful if you're trying to keep file size down and aren't worried about detecting other regions in the future.
1. Feel free to try it out with a bunch of different locales. The list of country codes provided in this PR is intended to be the official EU list. If you want more things to try, look at the unit tests for examples.

## Unit tests

So long as you have the npm modules installed, you can run the unit tests with a simple `npm test`.


lib/region/EU/countryCodes.js
```diff
@@ -0,0 +1,37 @@
...
+  LV: true, // Latvia
+  NL: true, // Netherlands
+  NO: true, // Norway
+  PL: true, // Poland
+  PT: true, // Portugal
+  RO: true, // Romania
+  SE: true, // Sweden
+  SI: true, // Slovenia
+  SK: true  // Slovakia
+};
```
aterranova-bv: Since these are all `true`, could we use an array here instead of an object?

In other words:
```js
var countryCodes: ['AT', 'BE']; // ... etc.
```

I think you gain readability, file size, and even [Performance](http://jsperf.com/jst-7) (if I set those tests up correctly).

taylormck: I believe the performance difference you were seeing was due to using Boolean coercion in one test and not the other. I edited the tests a bit to show both styles with and without boolean lookup. Map lookup is significantly faster than looking it up in an array. This is expected, as a lookup in a hash map is O(1), whereas lookup in a map is O(n).
http://jsperf.com/jst-7/2

aterranova-bv: Oh yeah, totally.  I was surprised that `Array.indexOf` was faster than the Boolean coerce to be honest, I originally had performance listed as a potential reason **not** to make the change.

But in our [lookup here](https://github.com/bazaarvoice/bv-ui-core/pull/51/files#diff-936ee39dcf72233a90d091daa51b6486R40) we're doing Boolean coerce so I was trying to compare apples with apples (`indexOf` would replace the coerce).

taylormck: So I used the `Boolean(foo)` style cast, which is pretty slow.
I went ahead and tried out the `!!` style coercion, which is apparently much faster - even faster than the lookup without the `!!`.
http://jsperf.com/jst-7/2 (Updated).

I'll go ahead and switch to that for performance reasons, though performance is hardly the main issue here. The main goal here is a function that takes in a locale and spits out a boolean. Returning an index instead would break the black box nature of this module.

aterranova-bv: Oh, nice.  Yeah, agreed on performance.

To keep the black boxness, I was thinking you would do 
```js
var isEULocale = countryCodes.indexOf(countryCode) !== -1;
```

Sorry if I wasn't clear on that.  I realize that means I wasn't actually comparing apples to apples :stuck_out_tongue_closed_eyes:

lib/region/EU/index.js
```diff
@@ -0,0 +1,43 @@
...
+ *   Returns false otherwise.
+ */
+module.exports = function (locale) {
+
+  // First, we need to fish out a country code. If the locale is of the xx_XX
+  // format, we need to toss the language part and keep the country code. If
+  // the locale is just a country code already, we can just keep it. Either
+  // way, we can just create split along '_' and grab the last one from the
+  // resulting list.
+  var countryCode = locale.split('_').pop();
```
tnunamak: There are web specs that use a hyphen as the standard for language/region separator (`ab-XY` instead of `ab_XY`). Can we support both?

lib/region/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+Loading the root module:
+```javascript
+var region = requrie('bv-ui-core/lib/region');
+
+region.isEULocale('en_US'); // false
+region.isEULocale('en_GB'); // true
+```
+This can be useful when you'll have to support detection of multiple regions.
+As we only support the EU region for now, it's not particularly useful yet.
```
aterranova-bv: Since this is a public repo, we may want to take this line out.

aterranova-bv:
:+1: , verification and tests successful.

@tnunamak 's comment is a good one.

lib/region/EU/index.js
```diff
@@ -0,0 +1,43 @@
...
+ *   Returns false otherwise.
+ */
+module.exports = function (locale) {
+
+  // First, we need to fish out a country code. If the locale is of the xx_XX
+  // format, we need to toss the language part and keep the country code. If
+  // the locale is just a country code already, we can just keep it. Either
+  // way, we can just create split along '_' and grab the last one from the
+  // resulting list.
+  var countryCode = locale.split(/[_\-]/).pop();
```
taylormck: @tnunamak Pretty clean change to support xx-XX locales, but it does involve adding a regex.

lib/region/README.md
taylormck: In this case, "we" means this particular module, not BV.
I'll go ahead and change it to an impersonal subject, though.

lib/region/EU/index.js
tnunamak: Works for me.

test/unit/region/EU/index.spec.js
```diff
@@ -0,0 +1,54 @@
...
+
+  it('should return true for IT', function () {
+    var result = isEULocale('IT');
+    expect(result).to.equal(true);
+  });
+
+  it('should return true for IT', function () {
+    var result = isEULocale('IT');
+    expect(result).to.equal(true);
+  });
```
KikoRamos: Is there a reason this test case was duplicated from the one above it?

taylormck: Nope, copypasta error.

taylormck: The extra test has been removed.

danheberden:
```<bikeshed>```

```js
region.match('en_GB', 'europe');

// region.js
var zones = {};
region.match = function(locale, zone) {
  if(!zones[zone]) {
     zones[zone] = require('./zones/' + zone);
  }
};

// OR

var euroRegion = require('region/europe');
euroRegion.contains('en_GB'); // true

```</bikeshed>```

**Comment by taylormck** [comment]
> ```javascript
> // region.js
> var zones = {};
> region.match = function(locale, zone) {
>   if(!zones[zone]) {
>      zones[zone] = require('./zones/' + zone);
>   }
> };
> ```

Dynamically requiring the modules still requires those modules to be packed & shipped, so you still don't get the advantage of smaller file size.

**Comment by danheberden** [comment]
Am i able to get the list of country codes manually as well?

**Comment by danheberden** [comment]
so 

```js
var europe = require('./lib/region/europe');
europe.contains('en_GB'); //true
```

would be better for size concerns?

taylormck:
> Am i able to get the list of country codes manually as well?

You can simply require the list.
```javascript
var europeanCountryCodes = require('bv-ui-core/lib/region/EU/countryCodes');
```

danheberden:
My mistake; i was looking at the files list and thought the var countryCodes = was part of the module, not its own file. :)

taylormck:
> so
> 
> ```javascript
> var europe = require('./lib/region/europe');
> europe.contains('en_GB'); //true
> ```
> would be better for size concerns?

That would eliminate the loading of other regions should we add them in the future.

taylormck:
If we'd prefer changing the `EU` to `europe`, I'm cool with that.
We can also change the `region` module to have a `match(locale, region)` method.

danheberden:
```js
var apac = require('./lib/region/apac');
apac.contains('ja_JP');

var europe = require('./lib/region/europe');
europe.contains('en_GB');
```

i'd argue about file size, though, because i'd want a comprehensive list. E.g., 

```js
var region = require('bv-ui-core/lib/region'); 
region.lookup('en_GB'); // prints out what region it is
region('europe'); // returns europe locales
region('europe').contains('en_GB'); // boolean for presence
```

danheberden:
since match implies RegExp, might be unsafe

danheberden:
I'm not trying to be a pain in the ass; just making hits the nail on the head if it gets into a bunch of people's projects

taylormck:
@danheberden I redid the API a bit. Have a look & tell me what you think (it's more or less what we talked about).

rmurphey:
Re dynamically requiring the modules, be sure to check out webpack's code splitting feature -- it would prevent the language modules from being in the main bundle, and would only load the one(s) needed.

lib/region/README.md
```diff
@@ -0,0 +1,38 @@
+# Region
+
+The `region` module is a module meant to standardize the detection of various
+regions, such as EU or APAC. It exposes functions that can test if a locale
```
tnunamak: I think APAC is an abbreviation that might need some definition, like `Asia-Pacific (APAC)`.

lib/region/README.md
```diff
@@ -0,0 +1,38 @@
...
+locale.
+
+So far, the only region supported is EU, but the module is designed to be
+easily extendable when the need to support other regions pops up.
+
+## Usage Examples
+
+Loading the root module:
+```javascript
+var region = requrie('bv-ui-core/lib/region');
```
tnunamak: Typo.

lib/region/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+region.has('en_US', 'europe'); // false
```
tnunamak: Maybe it's just me, but I think the API would read better like:

`region.has('europe', 'en_US')`
or
`region.is('en_US', 'europe')`

lib/region/index.js
```diff
@@ -0,0 +1,58 @@
...
+    // Loop through the regions looking for the locale.
+    var region;
+    for (region in regions) {
+      if (regions[region].has(locale)) {
+        return region;
+      }
+    }
+
+    // If we get here, the locale isn't in any currently supported locale.
+    return 'unsupported region';
```
tnunamak: It might be *slightly* more convenient (and semantic, IMO) if this was just `unsupported`.

Forgive me for being nitpicky!

lib/region/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+Loading a specific region module:
+```javascript
+var europe = require('bv-ui-core/lib/region/europe');
+
+europe.has('en_US'); // false
+europe.has('en_GB'); // true
+
+// Get an Array with all the region's country codes.
+var europeanCountryList = europe.listCountryCodes();
```
tnunamak: Further nitpickiness: I looked at [the spec](http://www.open-std.org/jtc1/sc22/wg20/docs/n610.pdf), and the language used indicates that a locale represents a "natural language" and a "territory". Since a territory is not necessarily a country, the use of "country" isn't completely semantically accurate.

I make no claim that you should change it, and I don't mean to build a bike shed; I'm particular just enough to share the findings :smiley:.

lib/region/europe/countryCodes.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * @file List of country codes legally considered part of the EU region.
+ */
+
+var countryCodes = {
```
tnunamak: I'm pretty sure I get it, but I'd still like to ask--what's the reasoning behind using this sort of data structure instead of a list?

lib/region/europe/index.js
```diff
@@ -0,0 +1,55 @@
...
+   *   Returns false otherwise.
+   */
+  has: function (locale) {
+
+    // First, we need to fish out a country code. If the locale is of the xx_XX
+    // format, we need to toss the language part and keep the country code. If
+    // the locale is just a country code already, we can just keep it. Either
+    // way, we can just create split along '_' and grab the last one from the
+    // resulting list.
+    var countryCode = locale.split(/[_\-]/).pop();
```
tnunamak: FWIW, I think if someone does `europe.has('fr')`, this returns `false`. This is technically correct, but the string `fr` parses into a countryCode of `fr`, which may have misleading side-effects if someone is not careful. It might be worth considering erroring out if the locale doesn't match the pattern.

lib/region/europe/countryCodes.js
taylormck: There was some discussion between @aterranova-bv and I around this that's getting hidden since I renamed this file (`EU` - > `europe`). Long story short, objects lookups are a very fast O(1), Array lookups are a slow O(n). This is because, in most javascript engines, Objects are hash maps while "Arrays" are actually linked lists. [Here's a performance test to show the difference](http://jsperf.com/jst-7/2).

Javascript Arrays certainly have their place. If you plan on modifying your list or iterating over it, Arrays are ideal. However, for constant lists that remain unchanged and are used only for lookup operations, Objects are king. Performance isn't really an issue for this module, but I still made it an object for common practice's sake.

lib/region/europe/index.js
taylormck: This is a good point; I'll toss in a `.toLowerCase` call.

tnunamak: Sorry, I should have elaborated more. A locale of just `fr` normally refers to the language, not the territory, in `fr_FR`. This API isn't intended to be used like `europe.has(language)`, it's about the territory.

That there is no territory included in the input is a sign that the user is misusing the API, and it might be better to alert them that they are doing something wrong instead of giving the technically correct, but wrong answer to the question they think they are asking:

```javascript
europe.has('fr') // false
europe.has('fr_FR') // true
```

The latter is supported, the former is unsupported but still returns a value.

lib/region/index.js
```diff
@@ -0,0 +1,58 @@
...
+ * be easy to add new regions, such as APAC, NA, LA, etc.
+ *
+ * This module is the "generic" module, and can be used to detect any region
+ * we support. If your project only needs to detect a specific region, you can
+ * require that region's module on its own to reduce file size.
+ *
+ * Examples:
+ *
+ * var region = require('bv-ui-core/lib/region');
+ * region.isEULocale('en_US'); // false
```
tnunamak: I don't see this API anywhere.

lib/region/europe/countryCodes.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * @file List of country codes legally considered part of the EU region.
```
tnunamak: Just curious, how'd you find these?

lib/region/europe/countryCodes.js
tnunamak: Thanks. I see it as a matter of personal taste, since you're right that for this module there are practically zero performance differences either way. Your reasoning is totally sound.

lib/region/index.js
taylormck: It was removed, but I missed this part of the comment.

tnunamak: And this causes verification steps to fail:

```
> region.isEULocale('en_US');
TypeError: region.isEULocale is not a function
    at repl:1:8
    at REPLServer.defaultEval (repl.js:164:27)
    at bound (domain.js:280:14)
    at REPLServer.runBound [as eval] (domain.js:293:12)
    at REPLServer.<anonymous> (repl.js:393:12)
    at emitOne (events.js:82:20)
    at REPLServer.emit (events.js:169:7)
    at REPLServer.Interface._onLine (readline.js:210:10)
    at REPLServer.Interface._line (readline.js:549:8)
    at REPLServer.Interface._ttyWrite (readline.js:826:14)
> region.has('en_US', 'europe')
false
> region.has('en_FR', 'europe')
```

lib/region/europe/countryCodes.js
taylormck: We had the BV legal team come up with a list and email it to me.

lib/region/index.js
taylormck: Yep, I'll update the PR, as well.

lib/region/europe/index.js
taylormck: Oh, I see. You're suggesting that we throw an error if the user provides us a lower case value.

danheberden: perhaps inference is dangerous here? `.hasCountry('fr')`, `hasLocale('fr_FR');`

taylormck: I like Dan's idea. Crystal clear interface, no we can implement it without a regex.

lib/region/europe/index.js
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+// territories is a map that contains an entry for each valid EU country
+// code, so territories[territory] will return either true or undefined.
+// We just need to coerce that into a boolean value and we're good to go.
+var territories = require('./territories');
+
+// Some constant patterns for valid values.
+var VALID_TERRITORY = /^[A-Z][A-Z]$/;
+var VALID_LOCALE = /^[a-z][a-z][\_\-][A-Z][A-Z]$/;
```
brianarn: Some basic simplifications:

```javascript
var VALID_TERRITORY = /^[A-Z]{2}$/;
var VALID_LOCALE = /^[a-z]{2}[-_][A-Z]{2}$/;
```

As regex goes, I feel like it makes them a bit clearer. The `[-_]` part works as `[_-]` as well, and neither character requires escaping. The `-` only has special meaning to imply ranges when it makes sense, like it does in `[a-z]`.

brianarn:
`lib/region/README.md` doesn't include any documentation that states that errors may be thrown if invalid input is supplied. When I was looking at the tests, I was surprised to see that errors were expected.

Having read the code a bit more, it makes sense now, and is totally a valid decision, but as a consumer of this code, I'd like documentation that clearly explains when intentional errors will be thrown, without requiring code review to be aware.

taylormck:
> Re dynamically requiring the modules, be sure to check out webpack's code splitting feature -- it would prevent the language modules from being in the main bundle, and would only load the one(s) needed.

I took a quick look at it, and it looks like we'd have to make all the methods in `region` asynchronous to use webpack's code splitting. For example:
```javascript
hasTerritory: function (region, territory) {

  var result;

  // Assume the module exists and the territory code is valid to keep this example simple.
  var regionPath = './' + region;

  // Asynchronously load the file so we can synchronously require it in the callback.
  require.ensure([regionPath], function (require) {
    var regionModule = require(regionPath);
    result = regionModule.hasTerritory(territory);
  });
 
  return result;
}
```

I feel like making the API asynchronous would be unintuitive, so I'm voting against it.
Still, if it's something we really want to do, I'll code it up.

danheberden:
I don't see anywhere that we're dynamically requiring modules. If you require the  main file, its dependencies are the regions. If you want, you can require just the region directly. something something keep it simple.

taylormck:
I believe @rmurphey's point was that we could have only the `region` module, then let it require the individual regions internally. Using webpack's code splitting feature, we could avoid loading the unused region's until they were needed. However, unless I've misunderstood something, it'd mean that we'd have to make the `region` module's API asynchronous.

tnunamak:
:+1:

I agree that requiring a region module directly is just fine.

PR #53, opened 2016-02-29
Add a static asset loader client.

reason-bv:
This adds an implementation of the static asset loader client to bv-ui-core, as a first step to opening the static asset service.

Note that this is a much simplified version of the first pass internal static asset service, as that had some issues with conflicting dependency declarations. Here dependencies are implicit in the request for resources: you request all of the dependencies you need with the asset.

Also included in this:

- catch up the changelog.
- minor logger changes.
- update old Node.js dependencies.


reason-bv:
@rmurphey Comments on the approach taken here appreciated.

lib/staticAssetLoader/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+The static asset loader must be configured with the following:
+
+  * A function to generate asset URLs from asset name and version strings.
+  * The name of the application namespace.
+
+Applications sharing static assets must use the same namespace.
+
+```javascript
+var namespacer = require('bv-ui-core/lib/namespacer');
```
tnunamak: Since the import of `namespacer` is shown, should it also be used in this example?

lib/staticAssetLoader/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+Note that when requesting an asset that has dependency assets - for example,
+Backbone depends on jQuery and either Underscore or Lodash - it is necessary to
+request those dependencies in the same `require` call.
+
+### Provide Assets
+
+Applications that ship with resources that might be useful to other applications
+on the page can provide those resources via the static asset loader. This again
+allows for a convergence of dependencies and reduced page weight.
```
tnunamak: I believe an asset can be loaded even if the corresponding `define()` call does not come until later in the execution path, right? Is this something worth being explicit about?

reason-bv: Yes, `require` and `define` can be called in any order. Though behavior becomes undefined if there is double definition of a given resource, as different people will have a different reference if the library itself doesn't prevent that. I should probably prevent that from happening in the loader code.

lib/staticAssetLoader/README.md
reason-bv: It should be removed; left over from a prior iteration.

lib/staticAssetLoader/README.md
```diff
@@ -0,0 +1,115 @@
...
+## Form of the Static Asset Service
+
+The static assert service is a web service that can respond to requests such as:
+
+`https://static.example.com/1/asset/namespace/jquery@1.11.1.js`
+
+Importantly, the static asset service must be capable of serving bundles of
+assets in response to a single request such as:
+
+`https://static.example.com/1/asset/namespace/jquery@1.11.1+backbone@1.2.0.js`
```
rmurphey: I may be recalling incorrectly, but I thought the service itself expected the requested modules to be in a specific (alphabetical??) order?

reason-bv: It does - there's a sort() in the URL generation code in this PR. I should correct this and note that alphabetical sorting is expected.

Issue #52, opened 2016-02-25
Prevent style sheets from loading twice


bv-jason-dippel:
There exists an issue where certain work has lead to situations where a single CSS file is being requested multiple times. This creates unnecessary load delays.

I propose we add a new module to bv-ui-core that provides the ability to check whether or not the CSS is already loaded and, if not, load a file from a given URL (if it's already present, just return). Ideally, we would have something that could be used as the following:

```js
import cssLoadChecker from 'bv-ui-core/lib/cssLoadChecker'

cssLoadChecker.checkAndLoadStyleSheet(namespace.cssUrl)
      .then( () => {
        // do something on success
      })
      .catch( (error) => {
        throw new Error('Error loading css: ' + error)
      })
```

Where `cssLoadChecker.checkAndLoadStyleSheet(namespace.cssUrl)` returns a promise that will be fulfilled upon loading/verifying the CSS file. 

Verifying will be done by checking for the existence of a certain style (ie display: none) on an empty div that we add with class `bv-verify-css-loaded` (or could possibly take in the class to look for as a param).

Something like this already exists within Firebird and is useful in the work I'm currently doing, and most likely future work as well. I believe this to be a good reason to add it to the bv-ui-core repo.

sirugh:
We would need to make sure that the case used to validate the css would work for any consumers. That may mean having a classname be passed as an option as well.

bv-jason-dippel:
Taking in the class name to look for is no problem. Determining what styles to look for as a means of knowing the CSS has loaded is a little more interesting imo. I don't think it's reasonable to take the styles in as a param as I believe it will unnecessarily complicate the code, but I'm curious as to what others think.

benhollander:
I haven't looked at the static asset service much - could it be used for this?

sirugh:
@benhollander AFAIK the SAS is used for loading less dynamic assets such as jQuery at a specific version, or constants that aren't going to change.

That being said and though client css is generated at build time, we may be able to modify the SAS to read from a bucket and direct requests for css files to the appropriate locations. Currently we just store the jquery/backbone/lodash code in the SAS repo. It would have to read from an external store to handle the css.

aterranova-bv:
> a single CSS file is being requested multiple times

Won't the browser cache the requested CSS file?

Or are we trying to solve multiple requests going out before the first one has returned?

bv-jason-dippel:
[...]

PR #55, opened 2016-03-09
Add css load checker to prevent duplicate loading of style sheet

bv-jason-dippel:
This adds the cssLoadChecker module, which is a means of preventing the duplicate loading of requested style sheets.

Research was done into the possibility of using the SAS client in bv-ui-core, but this would not work with the existing use case for this module as explained [here](https://github.com/bazaarvoice/bv-ui-core/issues/52#issuecomment-193389675).

### DIFF
--- a/lib/cssLoadChecker/README.md
+++ b/lib/cssLoadChecker/README.md
@@ -1,0 +1,19 @@
+# CSS Load Checker
+
+This module provides a means of checking if a CSS file has been loaded already or not. It does this by adding a test div to the page and checking the computed styles of it for the `display: none` style.
+
+```js
+var cssLoadChecker = require('bv-ui-core/lib/cssLoadChecker');
+
+if (cssLoadChecker.isCssLoaded('bv-verify-css-loaded')) {
+  // do stuff if loaded
+}
+else {
+  // do stuff if not loaded (like load it)
+  loader.loadStyleSheet(namespace.cssUrl, callback);
+}
+```
+
+The following methods are provided:
+
+- `isCssLoaded(className)`: Loads a test div onto the page with the given class name and checks for the style `display: none`. If this is set, then the CSS file is assumed to be present so it returns true. Otherwise, it returns false. The test div is removed before this function finishes executing.
--- a/lib/cssLoadChecker/index.js
+++ b/lib/cssLoadChecker/index.js
@@ -1,0 +1,72 @@
+/**
+ * @fileOverview
+ * Provides functions for checking if a stylesheet has been loaded previously
+ */
+
+// Imports
+var global = require('../global');
+var body = require('../body');
+
+/**
+ * Function to add the test div to the page
+ *
+ * @param {string} className - The class name used to check if the css file has loaded yet
+ *  (does not include the '.' out front, just the name)
+ * @return {object} The element that was just created for the purposes of
+ *  testing whether the CSS has loaded or not
+ */
+var _loadTestDiv = function (className) {
+  var testDiv;
+
+  testDiv = global.document.createElement('div');
+  testDiv.style.height = '0px';
+  testDiv.style.width = '0px';
+  testDiv.style.border = '0px';
+  testDiv.className = className;
+  body().appendChild(testDiv);
+
+  return testDiv;
+};
+
+/**
+ * Function to remove the test div from the page
+ *
+ * @param {object} testDiv - The element that was created for the purposes of
+ *  testing whether the CSS has loaded or not
+ */
+var _cleanupTestDiv = function (testDiv) {
+  testDiv.parentNode.removeChild(testDiv);
+};
+
+module.exports = {
+
+  /**
+   * Check the test div to see if the styles have loaded
+   *
+   * @param {string} className - The class name used to check if the css file has loaded yet
+   *  (does not include the '.' out front, just the name)
+   * @return {boolean} True or false as to whether the style sheet has been loaded.
+   */
+  isCssLoaded: function (className) {
+    var testDiv = _loadTestDiv(className);
+    var displayVal;
+    var result = false;
+
+    // add support for IE8
+    if (!global.getComputedStyle) {
+      displayVal = testDiv.currentStyle['display'];
+    }
+    else {
+      var computedStyles = global.getComputedStyle(testDiv);
+      displayVal = computedStyles.getPropertyValue('display');
+    }
+
+    if (displayVal === 'none') {
+      result = true;
+    }
+
+    _cleanupTestDiv(testDiv);
+    return result;
+  }
+
+};
--- a/lib/loader/README.md
+++ b/lib/loader/README.md
@@ -32,6 +32,10 @@ failure in old IE.
 attributes on the created `<script>` element.
 - **timeout**: the time in milliseconds before the loading should be
 considered failed. Defaults to 10000ms.
+- **forceLoad**: a boolean value as to whether the script should
+attempt to be loaded if an earlier attempt was made. Defaults to false.
+- **namespaceName**: a string containing the name of the namespace to check
+if the script has been previously loaded and store the loadedUrls hash
 
 ### loadStyleSheet options
 
@@ -41,6 +45,10 @@ attributes on the created `<link>` element.
 placed. The default insertion point is after the first `<script>` element on the page.
 - **timeout**: the time in milliseconds before the loading should be
 considered failed. Defaults to 10000ms.
+- **forceLoad**: a boolean value as to whether the stylesheet should
+attempt to be loaded if an earlier attempt was made. Defaults to false.
+- **namespaceName**: a string containing the name of the namespace to check
+if the stylesheet has been previously loaded and store the loadedUrls hash
 
 ### loader timeout
 
--- a/lib/loader/index.js
+++ b/lib/loader/index.js
@@ -14,9 +14,11 @@
  */
 
 var global = require('../global');
+var namespacer = require('../namespacer');
 var doc = global.document;
 
 var DEFAULT_TIMEOUT = 10000;
+var loadedUrls = {};
 
 function getFirstScript () {
   return doc.getElementsByTagName('script')[0];
@@ -36,6 +38,14 @@ function validateArguments (url, options, callback) {
     throw new Error('`url` must be a string');
   }
 
+  if (options.namespaceName && typeof options.namespaceName !== 'string') {
+    throw new Error('`options.namespaceName` must be a string');
+  }
+
+  if (options.forceLoad && typeof options.forceLoad !== 'boolean') {
+    throw new Error('`options.forceLoad` must be a boolean');
+  }
+
   if (typeof options.timeout !== 'number') {
     throw new Error('`options.timeout` must be a number');
   }
@@ -46,6 +56,27 @@ function validateArguments (url, options, callback) {
 }
 
 module.exports = {
+  /**
+   * Clears the loaded URLs
+   *
+   * Clears any URLs that might have been stored in the loadedUrls hash, either
+   * locally or in the given namespace(optional). This is needed for testing so
+   * we don't need unique URLs for every test, but rather this can be used as a
+   * means of cleaning up.
+   *
+   * @param {string} [namespace] - contains the name of the namespace to clear
+   *  the loadedUrls hash in.
+   */
+  _clearLoadedUrls: function (namespaceName) {
+    if (namespaceName) {
+      var namespace = namespacer.namespace(namespaceName);
+      namespace.loadedUrls = {};
+    }
+    else {
+      loadedUrls = {};
+    }
+  },
+
   /**
    * Load a script from a URL
    *
@@ -64,9 +95,17 @@ module.exports = {
    *  to be used as attributes on the created `script` element (optional)
    * @param {number} options.timeout - timeout in milliseconds (defaults to
    *  10000ms)
+   * @param {boolean} options.forceLoad - whether to load the script at the
+   *  specified url or not regardless of whether an earlier attempt to do so has
+   *  already been made
+   * @param {string} options.namespaceName - the name of the namespace to check
+   *  if the script has been previously loaded and store the loadedUrls hash
    * @param {function} [callback] - node-style callback
    */
   loadScript: function (url, options, callback) {
+    var _loadedUrls = loadedUrls;
+    var namespace;
+
     if (typeof options === 'function') {
       callback = options;
       options = null;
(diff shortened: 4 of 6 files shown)


lib/cssLoadChecker/index.js
```diff
@@ -0,0 +1,103 @@
...
+ * Function to check whether the className corresponds to
+ * a loaded style sheet
+ *
+ * @param {string} className - The class name used to check if the css file has loaded yet
+ *  (does not include the '.' out front, just the name)
+ * @return {boolean} True or false as to whether the style sheet has been loaded.
+ */
+var _checkForCss = function (className) {
+  var testDiv = _loadTestDiv(className);
+  var computedStyles = global.getComputedStyle(testDiv);
```
reason-bv: Not supported in IE8, so I think some kind of check is needed here.

lib/cssLoadChecker/index.js
```diff
@@ -0,0 +1,103 @@
...
+ */
+var _loadTestDiv = function (className) {
+  var testDiv;
+
+  testDiv = global.document.createElement('div');
+  testDiv.style.height = '0px';
+  testDiv.style.width = '0px';
+  testDiv.style.border = '0px';
+  testDiv.className = className;
+  global.document.body.appendChild(testDiv);
```
reason-bv: There is a `body` module in `bv-ui-core`. Better to use that.

reason-bv:
So it seems to me that rather than doing what you are doing here, you'd be better off extending the loader to keep a record of what it loads, and pass in a flag to force reload, and just make sure it never tries to load stylesheets twice by default.

If you do have other use cases for checking computed styles to see what they are, then salvaging the core of this may still be helpful for other uses.

sirugh:
We should at least keep the checker module separate from the loader. Requesting and parsing CSS are two separate concerns and should be captured here in this PR.

tnunamak:
> So it seems to me that rather than doing what you are doing here, you'd be better off extending the loader to keep a record of what it loads, and pass in a flag to force reload, and just make sure it never tries to load stylesheets twice by default.

The PRS use case driving this involves the PRS waiting for a stylesheet that Firebird is responsible for loading. Unless multiple instances of the loader share state about what is loaded, we'll still need a way to recognize when CSS that we did not load is available.

sirugh:
@tnunamak I do see what @reason-bv was saying though. There is already functionality to issue requests for resources, so part of this PR is unnecessary.

tnunamak:
Lack of due diligence on my part, you're right.

reason-bv:
So in theory we're all using a shared namespace, so there could be a trackedloader module that writes to that namespace, or the loader could take a configuration that tells it to write to a shared place in that namespace. We should be all about sharing state between apps when it helps us.

bv-jason-dippel:
So in response to @reason-bv's comments, I've added some stuff to the loader module that keeps track of what URLs we've loaded files from and will only attempt to reload them if the options.forceLoad flag is passed.

I do believe we should keep the cssLoadChecker module in addition to this for the specific use case that brought these changes around. The PRS use case driving this does not have access to the URL firebird uses to load the CSS, so we must check whether the CSS has loaded yet or not by looking at the computed styles.

lib/cssLoadChecker/index.js
```diff
@@ -0,0 +1,73 @@
...
+
+    if (displayVal === 'none') {
+      result = true;
+    }
+
+    _cleanupTestDiv(testDiv);
+    return result;
+  }
+
+}
```
reason-bv: Semicolon.

lib/loader/index.js
```diff
@@ -14,9 +14,13 @@
  */
 
 var global = require('../global');
+var namespacer = require('../namespacer');
 var doc = global.document;
 
 var DEFAULT_TIMEOUT = 10000;
+var SHARED_NAMESPACE_NAME = 'BV_LOADER_TRACKING';
```
reason-bv: I was thinking you'd take an optional namespace as a parameter / setting / something, so that only apps sharing a namespace share this.

For CSS it doesn't matter, but for JS we have the interesting situation in which app in namespace A and app in namespace B may both be using the static asset service to load the same content into different encapsulations in different namespaces.

This approach could mess that up. Thoughts from the other folk?

bv-jason-dippel: How would you feel about defaulting to SHARED_NAMESPACE_NAME, but allow users the option of passing in a specific namespace? Best of both worlds?

reason-bv: Since this is open and not BV-specific, I think I'd rather see it default to not adding to a namespace. E.g. keep a local reference to a hash object in loadedUrls, and only add it to a namespace if provided with one.

lib/loader/index.js
```diff
@@ -14,9 +14,13 @@
...
+var namespace = namespacer.namespace(SHARED_NAMESPACE_NAME);
+namespace.loadedUrls = namespace.loadedUrls || new Array();
```
reason-bv: Why `new Array()` and not `[]`?

lib/cssLoadChecker/index.js
```diff
@@ -0,0 +1,73 @@
+/**
+ * @fileOverview
+ * Provides functions for checking if style sheet is already loaded. May attempt to load the style
+ * sheet if not present.
```
reason-bv: This is no longer true.

lib/loader/index.js
```diff
@@ -36,6 +38,14 @@ function validateArguments (url, options, callback) {
     throw new Error('`url` must be a string');
   }
 
+  if (options.namespace && typeof options.namespace !== 'string') {
```
reason-bv: To remove confusion, call this namespaceName, perhaps?

lib/loader/index.js
```diff
@@ -77,6 +113,24 @@ module.exports = {
 
     validateArguments(url, options, callback);
 
+    if (options.namespace) {
+      namespace = namespacer.namespace(options.namespace);
```
reason-bv: Better to get a static reference to the namespace early than to run through the function each time. It's a small thing here, since this won't be called frequently. Still.

lib/loader/index.js
```diff
@@ -155,6 +217,9 @@ module.exports = {
    *  to be used as attributes on the created `link` element (optional)
    * @param {number} options.timeout - timeout in milliseconds (defaults to
    *  1000ms)
+   * @param {boolean} options.forceLoad - whether to load the style sheet at the
+   *  specified url or not regardless of whether an earlier attempt to do so has
+   *  already been made
```
reason-bv: Need to add the namespace to the options.

lib/loader/index.js
```diff
@@ -235,6 +318,13 @@ module.exports = {
 
       try {
         injectionNode.appendChild(link);
+
+        if (options.namespace) {
```
reason-bv: You could probably simplify these sorts of if-block by assigning loadedUrls to namespace (and reassign when clearing), so you'd always be able to check the local variable.

reason-bv:
Ok, seems headed in the right direction. Comments made on the latest are all I have, the big picture other than those is ok.

lib/loader/index.js
bv-jason-dippel: The reason I put it there as opposed to the top of the function was so that I could include the validation logic within the `validateArguments` function. To me it makes sense to group all that together in the same function, even if it means I get the namespace a little later. If you still disagree and feel like it should be moved to the top of the function, let me know and I'll update it.

lib/loader/README.md
```diff
@@ -41,6 +45,10 @@ attributes on the created `<link>` element.
 placed. The default insertion point is after the first `<script>` element on the page.
 - **timeout**: the time in milliseconds before the loading should be
 considered failed. Defaults to 10000ms.
+- **forceLoad**: a boolean value as to whether the script should
```
sirugh: s/script/stylesheet

lib/loader/index.js
```diff
@@ -47,6 +57,27 @@ function validateArguments (url, options, callback) {
...
+   * we don't need unique URLs for every test, but rather this can be used as a
+   * means of cleaning up.
+   *
+   * @param {string} [namespace] - contains the name of the namespace to clear
+   *  the loadedUrls hash in.
+   */
+  clearLoadedUrls: function (namespaceName) {
+    if (namespaceName) {
+      var namespace = namespacer.namespace(namespaceName);
+      namespace.loadedUrls = [];
```
sirugh: Using namespacer, the proper syntax would be something like

```
try {
 namespace.registerProperty('loadedUrls', []);
} catch (e) {
 // already registered, probably log and do nothing.
}
```

lib/loader/index.js
```diff
@@ -135,6 +186,16 @@ module.exports = {
       var firstScript = getFirstScript();
 
       firstScript.parentNode.insertBefore(script, firstScript);
+
+      _loadedUrls[url] = true;
+
+      if (options.namespaceName) {
+        namespace.loadedUrls = _loadedUrls;
```
reason-bv: Though @sirugh has a point, you should use namespace.registerProperty when first setting a property on the namespace, as the fact that this throws to prevent overwriting is important.

test/unit/loader/index.spec.js
```diff
@@ -197,6 +218,79 @@ describe('lib/loader', function () {
...
+        scriptExecuted = 1;
+      };
+
+      loader.loadScript(url, {
+        namespaceName: TESTING_NAMESPACE_NAME
+      }, function () {
+        expect(scriptExecuted).to.equal(1);
+        expect(namespace.loadedUrls[url]).to.equal(true);
+        done();
+      });
```
sirugh: This function should attempt to load the same url a second time and assert that it is the same as the loaded script from the first execution of `loadScript`.

sirugh: The same goes for the stylesheet check in the same vein.

lib/loader/index.js
```diff
@@ -47,6 +57,27 @@ function validateArguments (url, options, callback) {
 
 module.exports = {
   /**
+   * Clears the loaded URLs
+   *
+   * Clears any URLs that might have been stored in the loadedUrls hash, either
+   * locally or in the given namespace(optional). This is needed for testing so
```
sirugh: If the only reason this exists is for testing, it probably doesn't need to exist, at least not in the module. Is there another way to get a handle to an internal data structure and clear it via tests? You can get a handle on namespace.loadedUrls but not the internal loadedUrls right?

lib/loader/index.js
sirugh: I deleted that comment because I realized it was not relevant on the given function, but yes.

lib/loader/index.js
bv-jason-dippel: That is correct. During testing, I'm able to access namespace.loadedUrls but not the internal loadedUrls because I'm not exporting it. The best way I've thought of doing it is with this function, but as you said if it's only purpose is for testing, it would be nice to remove it.

I don't think there's a need for it in actual execution, but I've not had any luck removing it and still being able to test things nicely. I'm open to suggestions.

reason-bv: I'm not adverse to exposing things for test purposes.

sirugh: If you absolutely must, then I suggest marking it as private `_clearLoadedUrls` or `_cleanup` or something that indicates it's not part of the public API of this module.

lib/loader/index.js
bv-jason-dippel: @sirugh, @reason-bv Just to be clear, you guys are talking about something like...
```js
_loadedUrls[url] = true;

if (options.namespaceName) {
  if (namespace.loadedUrls) {
    namespace.loadedUrls = _loadedUrls;
  }
  else {
    namespace.registerProperty('loadedUrls', _loadedUrls);
  }
}
else {
  loadedUrls = _loadedUrls;
}
```

reason-bv: Not quite. The first time you set that property on the namespace, you have to register it. Not afterwards.

So it might be better for you to, at the outset and once, register an object on the namespace, then add loadedUrls to that object as you need to elsewhere in the code.

tnunamak: As I understand it, by calling `registerProperty`, you are establishing ownership over that property. Multiple apps can read it, but only one should call `registerProperty` and that is the only one that should be allowed to modify it. Correct?

sirugh: They can de-reference it or assign properties of the property. `registerProperty` is just a helper that throws an error if it happens to exist already. So if you are performing possible destructive ops you can use it to be safe.

tnunamak: I've never used [`set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set) before, could we use it to enforce this? Could you determine if the property owner is the one making a modification by inspecting the context?

tnunamak: > They can de-reference it or assign properties of the property.

Assign properties as in modify its value? This seems like a risky thing to allow.

reason-bv: What @sirugh said - use of registerProperty is just a way to ensure that when there are conflicts, there will be a clear exception rather than confusion as two apps edit each other's data haphazardly.

I don't think there is any intent of having app2 read the data used by app1.

tnunamak: Okay. As long as we're not advocating for namespace properties to be shared, this seems fine. If/when we do want apps to share state, we will want to use namespace somehow, won't we?

reason-bv: Yes. 

In fact what is going on here is sharing of properties between the loaders of different apps using the same namespace.

Which means I'm being dumb about what I'm suggesting with registerProperty. The same code running against the same namespace in different apps was one of the original goals here. So hmm. Don't use register property. Do write some inline documentation to explain why you are not using it.

lib/loader/index.js
taylormck: It's possible to write tests that would require this, but as it's just a "clean" method, it's fine by me to have around.

sirugh: Then I'm fine having it `_clearLoadedUrls`.

lib/cssLoadChecker/README.md
```diff
@@ -0,0 +1,19 @@
+# CSS Load Checker
+
+This module provides a means of checking if a CSS file has been loaded already or not. It does this by adding a test div to the page and checking the computed styles of it for the `display: none` style.
+
+```js
+var cssLoadChecker = require('bv-ui-core/lib/cssLoadChecker');
+
+if (cssLoadChecker.checkForCss('bv-verify-css-loaded')) {
```
sirugh: 1. Should this be a promise? Technically the promise only needs to be resolved once as once css loads, it's "loaded" and can't be unloaded. Am I just overengineering? @savetheclocktower? @tnunamak?

```javascript
cssLoadChecker.cssIsLoaded('bv-verify-css-loaded')
  .then(() => { 
    // Continue with your life.
  }
  .fail(() => {
     loader.loadStyleSheet(namespace.cssUrl, callback)
  })
```

2. This is personal preference: `checkForCss` -> `isCssLoaded` since it returns a boolean.

tnunamak: If it's not an async check, using a promise does feel like overengineering to me.

reason-bv: Promises would require a library, and we're trying to keep to zero dependencies here.

sirugh: Oh. Right. Duh. Thanks for the input guys. Carry on.

lib/loader/index.js
```diff
@@ -77,6 +116,26 @@ module.exports = {
...
+     * The goal of using the namespacer is to allow multiple apps on the same
+     * page to share the properties within the namespace, namely loadedUrls. For
+     * this reason, we are forgoing the use of registerProperty to instantiate
+     * the list of loaded URLs. This will allow the loader module to be run in
+     * different apps but access the same namespace property.
+     */
+
+    if (options.namespaceName) {
+      namespace = namespacer.namespace(options.namespaceName);
+      _loadedUrls = namespace.loadedUrls || [];
```
sirugh: Default to object. Cannot reference array index by string.

lib/loader/index.js
```diff
@@ -235,6 +332,15 @@ module.exports = {
 
       try {
         injectionNode.appendChild(link);
+
```
sirugh: If `appendChild` and the resulting script inject is synchronous, this is fine, but if it isn't you may consider putting 336-343 in the `onload`.

sirugh:
LGTM.

PR #60, opened 2016-07-18
UIA-8022 - Changing domain police to allow 'www.' with any url hostname

mgoodnow:
…ut the need to explicitly define

### DIFF
--- a/lib/domainPolice/README.md
+++ b/lib/domainPolice/README.md
@@ -48,6 +48,8 @@ nopeCop.get('anything'); // => undefined
 for later lookup.
 
 Prefixing `.` on the `domain` will allow it to match all subdomains as well.
+URL hosts which include `www.` will be allowed with domain name match.
+
 Given the above example, the following domains are all valid:
 
 - `bazaarvoice.com`
@@ -55,6 +57,7 @@ Given the above example, the following domains are all valid:
 - `localhost`
 - `www-stg.localhost`
 - `no-prefixing-dot.foo.com`
+- `www.no-prefixing-dot.foo.com`
 
 Domains that would not be allowed:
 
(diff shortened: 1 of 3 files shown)


sirugh:
Please update or add [unit tests](https://github.com/mgoodnow/bv-ui-core/blob/master/test/unit/domainPolice/index.spec.js).

mgoodnow:
Unit tests updated

sirugh:
Failing due to phantomjs rate limit....

aterranova-bv:
Is it useful to be able to set your preferred policy (all sub-domains, or not) when constructing a domain police object?

In other words, this change is going to alter functionality for existing users. Should we provide something for backwards compatibility?

reason-bv:
Allowing subdomains by default is a security threat in the abstract. Consider the issues that GitHub had with that, for example. It isn't necessarily the case that a company would want by default all subdomains to be allowable.

mgoodnow:
@reason-bv not sure I'm familiar with the security concerns you are raising, but I'll trust your judgement. Currently confighub does not store invalid domain (due to them being used for other purposes), thus allowing user to enter in `.domain.com` would be considered invalid. One potential option we could do is how we handle `allowThirdParty` in that we have a checkbox for `allowSubDomains`. I'll run this by Patrick T. Thanks for your feedback.

sirugh:
https://github.com/blog/1452-new-github-pages-domain-github-io

mgoodnow:
@reason-bv and @aterranova-bv Please review latest comment in https://bits.bazaarvoice.com/jira/browse/CCS-6339 for potential aternative measures at solving the problem at hand. Thanks.

mgoodnow:
Updated PR to include allowing `www.` along with any hostname.

taylormck:
I'm 👍  with these changes.
I'd still like to ping @reason-bv for a final look before merging.

reason-bv:
I'm good with this.

PR #61, opened 2016-08-15
JST-79: check for callback existence before invoking

sirugh:
We were unsafely calling `callback` on any subsequent loads after the first of a url.

ezpz check that we just missed.

Also added unit tests that fail in phantom but pass in chrome for some reason. *shrug*

### DIFF
--- a/lib/loader/index.js
+++ b/lib/loader/index.js
@@ -132,7 +132,9 @@ module.exports = {
     if (!options.forceLoad && _loadedUrls[url]) {
       // File at given url has already been loaded and option has not been set
       // to reload it, so we'll just call the callback and return.
-      callback();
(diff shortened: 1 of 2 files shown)


lib/loader/index.js
```diff
@@ -132,7 +132,9 @@ module.exports = {
     if (!options.forceLoad && _loadedUrls[url]) {
       // File at given url has already been loaded and option has not been set
       // to reload it, so we'll just call the callback and return.
-      callback();
+      if (typeof callback === 'function') {
```
Urthen: What happens if Javascript redefines the string constant representing the type of a function object, HUH?????

Urthen:
👍

PR #63, opened 2016-08-22
JST-80: Add an API module for communicating with the bazaarvoice api.

sirugh:
This adds a module that allows easy interaction with the bazaarvoice API.

Each request type should return a promise that resolves with the response from the API.

You can interact with the api module by using the `.get` syntax or you can import the specific module type you want:

```javascript
import api from 'bv-ui-core/api'
api.get('statistics', options).then(...)
```

```javascript
import getStatistics from 'bv-ui-core/api/statistics'
getStatistics(options).then(...)
```

Currently this only adds a `statistics` request type.

Also note that in the API [readme](https://github.com/bazaarvoice/bv-ui-core/pull/63/files#diff-9174f271f80a9817b3ee2debd1074a86R6) I suggested that people either require this source and then transpile or use our `dist/` folder which would contain the transpiled code. I have not added any build/deploy process to create that dist folder but I can. I just wanted to see how this felt to people.


lib/domainPolice/index.js
```diff
@@ -25,7 +25,7 @@ function domainMatches (domain, host) {
     return index >= 0 && index === (1 + host.length - domain.length);
   }
 
-  return host === domain || host === ("www." + domain);
+  return host === domain || host === ('www.' + domain);
```
taylormck: Either use semicolons or not, but pick one.

lib/api/statistics.js
```diff
@@ -0,0 +1,129 @@
...
+ * @param  {Array}  options.key - client's api key
+ * @param  {string}  options.type - Using "Reviews" returns statistics for all
+ *         content, including syndicated content (if enabled on your API key).
+ *         If you only want statistics for reviews you own that were written for
+ *         the products specified, use "NativeReviews" instead.
+ * @param  {Object}  [options.filters] - Filters object. Keyed by filter name.
+ * @return {Promise} a promise that will be resolved with the raw results
+ *                   from the statistics call.
+ */
+function getStatistics (options) {
```
taylormck: Just a thought: since we're using ES6 now, you can use destructuring to make the function signature more explicit. Then we can just let the Javascript engine do this parameter checking for us.
```javascript
function getStatistics ({ productIds, environment, key, type, filters: {} })
```

lib/domainPolice/index.js
sirugh: I was fixing a linting issue with double-quotes. Semicolons don't have a project-wide rule. This module uses semicolons. So no, I will not make that decision for this module.

lib/api/statistics.js
sirugh: Good suggestion. Also `filters = {}` for destructuring.

karma.conf.js
```diff
@@ -45,7 +35,10 @@ module.exports = function (config) {
         served: true,
         watched: true,
         nocache: true
-      }
+      },
+
+      // We need the polyfill for testin es6 modules.
```
aterranova-bv: s/testin/testing

taylormck: I would also accept `testin'`.

lib/api/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+This module acts as a javascript wrapper around api.bazaarvoice.com. For more
+information about the API, please see the [API documentation][0].
+
+### Requirements
+
+This module is written using ES6. In order to use the source you may need to
+provide your own transpiler, depending on your supported browsers.
+
+We also takes advantage of some built-ins that you may need to polyfill:
```
aterranova-bv: s/takes/take

lib/api/README.md
```diff
@@ -0,0 +1,130 @@
...
+#### How to Use
+
+Before using, you'll need to load up the API module:
+
+```javascript
+const api = require('bv-ui-core/lib/api')
+```
+
+Or if you know the specific type of request you are going to make you can
+call, you can do the following:
```
aterranova-bv: The last part of this sentence is confusing.

taylormck: Double predicate ?!
What does it mean ?!

lib/api/README.md
```diff
@@ -0,0 +1,129 @@
...
+api.get('statistics', options)
+
+// or
+const getStatistics = require('bv-ui-core/lib/api/statistics')
+getStatistics(options)
+```
+
+#### Parameters
+
+##### options
```
aterranova-bv: This is awesome.  Do you have any good strategies for keeping this in sync with the code? Like what happens when something gets added to `options` - how does a developer know that they should update this also?

lib/api/index.js
```diff
@@ -0,0 +1,15 @@
+import { default as getStatistics } from './statistics'
+
+const typeMap = {
+  statistics: getStatistics
+}
+
+function get (type, options = {}) {
```
aterranova-bv: Nice

lib/api/statistics.js
```diff
@@ -0,0 +1,117 @@
...
+ * @param  {string}  options.type - Using "Reviews" returns statistics for all
+ *         content, including syndicated content (if enabled on your API key).
+ *         If you only want statistics for reviews you own that were written for
+ *         the products specified, use "NativeReviews" instead.
+ * @param  {Object}  [options.filters] - Filters object. Keyed by filter name.
+ * @return {Promise} a promise that will be resolved with the raw results
+ *                   from the statistics call.
+ */
+function getStatistics ({ productIds, environment, key, type, filters = {} }) {
+  if (!productIds) {
```
aterranova-bv: `!Array.isArray(productIds)` ?

lib/api/statistics.js
```diff
@@ -0,0 +1,117 @@
...
+    throw new TypeError('productIds must be an array')
+  }
+  if (!environment) {
+    throw new TypeError(
+      'environment must be \'qa\', \'staging\', or \'production\''
```
aterranova-bv: If a consumer passes in 'prod' or 'stg' unwittingly, will the promise ever get rejected?

lib/api/statistics.js
```diff
@@ -0,0 +1,117 @@
...
+    )
+  }
+  if (!key) {
+    throw new TypeError('key must be provided')
+  }
+  if (!type) {
+    throw new TypeError('type must be \'Reviews\' or \'NativeReviews\'')
```
aterranova-bv: Same question as above

lib/api/statistics.js
```diff
@@ -0,0 +1,117 @@
...
+  const uri = `https://${envPrefix}.bazaarvoice.com/data/statistics.json` +
+            `?apiversion=${API_VERSION}` +
+            `&passkey=${key}` +
+            `&stats=${type}` +
+            Object.keys(filters)
+              .map(filter => `&filter=${filter}:${filters[filter]}`)
+              .join()
+
+  // Clone the productIds so we can manipulate it.
+  productIds = productIds.slice()
```
aterranova-bv: I'd never done this before - interesting.  While researching I also found that ES6's new spread (`...`) operator can be used to do the same thing.  Check out the "A more powerful array literal" section in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator.

I don't know that I prefer one over the other, just wanted to share.

aterranova-bv: Actually I'll just paste that section to make it easy:

> A more powerful array literal

> Example: Today if you have an array and want to create a new array with the existing one being part of it, the array literal syntax is no longer sufficient and you have to fall back to imperative code, using a combination of push, splice, concat, etc. With spread syntax this becomes much more succinct:

```js
var parts = ['shoulders', 'knees'];
var lyrics = ['head', ...parts, 'and', 'toes']; // ["head", "shoulders", "knees", "and", "toes"]
```

test/unit/api/index.spec.js
```diff
@@ -0,0 +1,33 @@
...
+        environment: 'qa',
+        key: 'clients_api_key',
+        type: 'Reviews',
+        filters: {
+          ContentLocale: 'en_US'
+        }
+      })
+    })
+  })
+})
```
aterranova-bv: Along the lines of one of my previous comments, we may want to test some failure scenarios - mainly that the Promise rejects properly if an invalid URL is built based on the options passed in.

lib/api/statistics.js
sirugh: Yes, but I should probably add some extra validation here.

```javascript
const envPrefix = envPrefixMap[environment]
if (!envPrefix) {...}
```

aterranova-bv:
👍

lib/api/README.md
taylormck: Just a note: https://bits.bazaarvoice.com/jira/browse/VJB-729

PR #66, opened 2017-03-10
Prevent namespaces from being replaced or removed

tnunamak:
Currently, namespaces [cannot replace existing namespaces](https://github.com/bazaarvoice/bv-ui-core/blob/master/lib/namespacer/index.js#L88). I'd like to propose that we disallow namespaces being replaced by different values in general.

```javascript
> window.BV
{ a: 1 }
> window.BV.b = 2  // the namespace is extensible
{ a: 1, b: 2 }
> window.BV = { c: 3 }
> window.BV        // but not writable
{ a: 1, b: 2 }     
```

The motivation for this is that my namespace is shared by different runtime consumers, and certain properties on the namespace are guaranteed to be read-only. A consumer could work around this by replacing the entire namespace with writeable properties, which I want to disallow.

P.S. This seems to be a tricky unit test to write. Something in the test runner doesn't seem to be respecting the property descriptor, as writes appear to be allowed anyway, even when the code under test throws an error if modifications to `global[name]` are attempted.

### DIFF
--- a/lib/namespacer/index.js
+++ b/lib/namespacer/index.js
@@ -64,7 +64,9 @@ module.exports = {
     // If there is no global property yet assigned for this namespace then
     // create one.
     if (global[name] === undefined) {
-      global[name] = new Namespace(name);
+      Object.defineProperty(global, name, {
+        value: new Namespace(name)
+      });
     }
     // If an object already exists for this global property, check to see
     // whether or not it is a Namespace instance or some other form of object.


reason-bv:
What is the use case requiring non-writable properties in this case?

A client or malicious third party Javascript can always break things; you can't really stop that with non-writable properties.

sirugh:
> What is the use case requiring non-writable properties in this case?

Disregarding bad actors, we just want a simple gate against accidental manipulation by our own applications.

Velveeta:
^^ That, I was gonna say, there's typically always a way for malicious users to do what they want, but we should at least have some rudimentary blocks for the accidental overwrite, which should signal to devs that this is a no-no.

tnunamak:
Any ideas as to why Travis is failing? @sirugh it looks like your last PR had the same issue?

```bash
$ npm test

> bv-ui-core@0.11.0 test /home/travis/build/bazaarvoice/bv-ui-core
> ./node_modules/karma/bin/karma start --single-run --browsers PhantomJS

Hash: f4683f5fa2953dc3a97c
Version: webpack 1.14.0
Time: 164ms

ERROR in ./test/unit/mochaInit.js
Module build failed: /home/travis/build/bazaarvoice/bv-ui-core/node_modules/babel-plugin-lodash/lib/index.js:85
    Program(path, state) {
           ^
SyntaxError: Unexpected token (
    at exports.runInThisContext (vm.js:73:16)
    at Module._compile (module.js:443:25)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at /home/travis/build/bazaarvoice/bv-ui-core/node_modules/babel-core/lib/transformation/file/options/option-manager.js:178:20
    at Array.map (native)
    at Function.normalisePlugins (/home/travis/build/bazaarvoice/bv-ui-core/node_modules/babel-core/lib/transformation/file/options/option-manager.js:158:20)
    at OptionManager.mergeOptions (/home/travis/build/bazaarvoice/bv-ui-core/node_modules/babel-core/lib/transformation/file/options/option-manager.js:234:36)
    at OptionManager.init (/home/travis/build/bazaarvoice/bv-ui-core/node_modules/babel-core/lib/transformation/file/options/option-manager.js:368:12)
    at File.initOptions (/home/travis/build/bazaarvoice/bv-ui-core/node_modules/babel-core/lib/transformation/file/index.js:212:65)
    at new File (/home/travis/build/bazaarvoice/bv-ui-core/node_modules/babel-core/lib/transformation/file/index.js:135:24)
    at Pipeline.transform (/home/travis/build/bazaarvoice/bv-ui-core/node_modules/babel-core/lib/transformation/pipeline.js:46:16)
    at transpile (/home/travis/build/bazaarvoice/bv-ui-core/node_modules/babel-loader/lib/index.js:46:20)
```

sirugh:
@tnunamak no idea. Try locally? Same version of webpack/node/etc?

tnunamak:
Looks like master no longer works on 0.12.18, which is what Travis is configured to run tests with.

tnunamak:
I'll rebase when #67 merges.

tnunamak:
It's passing now.

sirugh:
Add a unit test and I'll mergey merge.

PR #67, opened 2017-04-11
Use latest Node

tnunamak:
[Tests are breaking](https://travis-ci.org/bazaarvoice/bv-ui-core/builds/209881734) with 0.12.18. This is reproducible on master.

This tells Travis to use latest, and documents it in package.json for users.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: node_js
 node_js:
-  - "0.12"
+  - "7.8"
 install:
   - npm install
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: node_js
 node_js:
-  - "7.8"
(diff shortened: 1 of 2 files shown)


sirugh:
Probably should be using LTS, 6.10.2.

Velveeta:
FYI, I'm all for using the latest and greatest, but when it comes to Node, it's typically advised not to use the odd-numbered major releases in any kind of production capacity. You generally use them to do feature exploration for the next upcoming even-numbered major release, or to start testing your code's compatibility with upcoming changes to ensure you're not gonna have a major refactor when the next even-numbered major release drops. The odd-numbered releases are always considered experimental.

tnunamak:
That's fine, v6 (LTS) works, and I've updated this to use it.

PR #69, opened 2017-05-02
SWAT-1020 -- Adding new 'ready' function/Promise interface

Velveeta:
[SWAT-1020](https://bits.bazaarvoice.com/jira/browse/SWAT-1020)

Per a discussion yesterday, it would be nice to be able to queue up callbacks for application classes, which wouldn't fire until the application was deemed 'loaded' and available for interacting with. The intent is that applications loaded via BV Loader should automatically kick off the callback queue once they're finished being loaded.

After some discussing about the interface, we came to an agreement that in environments where ES6 Promises were not available, `ready` will behave as a simple callback aggregator, and return undefined: e.g. `app.ready(callback1); app.ready(callbackN);`. In environments where Promises *are* available, either due to native support, or polyfilled support, it will provide the same callback aggregation, but also return a Promise that can be used to queue up more callbacks (or the Promise interface could be used exclusively), e.g. `app.ready(callback1).then(callback2); app.ready().then(callbackN);`

Callbacks will be given an instance of the loaded application, and when passed to the `ready` function itself, should follow a standard Node errback function signature (err, app), versus when passed via a Promise chain, should split those parameters for their `then` and `catch` branches, e.g. `app.ready((err, app) => {})` or `app.ready().then((app) => {}).catch((err) => {})`.

### DIFF
--- a/lib/application/README.md
+++ b/lib/application/README.md
@@ -31,29 +31,33 @@ namespacer.namespace('APP').registerProperty(
 // loader.load(url).then( ... do something ... );
 ```
 
+## ES6 Considerations
+
+This module takes advantage of Promises, which you may need to [polyfill][3]:
+
+- `Promise`: http://caniuse.com/#feat=promises
+
 ## Methods to be Called by Users
 
 An application instance provides the following methods:
 
-- `ready(callback)`: A method intended for your users to call at any time in
-  order to register a callback which will be invoked as soon as your
-  application has been loaded and its API is available to be used. Callback
-  function should be a standard Node errback signature, where its first
-  expected parameter will be an error (if applicable), and its second
-  parameter will be the application instance.
+- `ready(callback)`: (optional) A method intended for your users to call at
+  any time in order to register a callback which will be invoked as soon as
+  your application has been loaded and its API is available to be used.
+  Callback function should be a standard Node errback signature, where its
+  first expected parameter will be an error (if applicable), and its second
+  parameter will be the application instance. The return value is a Promise
+  that you can chain your callbacks onto. The success callbacks (on your
+  `then` Promise chain) will only be given a single input parameter, the
+  application instance, and your failure callbacks (on your `catch` Promise
+  chain) will only be given a single input parameter,
+  the error.
 
-  Note that the callback itself is optional in environments where ES6 Promises
-  are available for use. In these scenarios, you can use the return value of
-  `ready` as a Promise to chain your callbacks onto. In this case, your
-  success callbacks (on your `then` Promise chain) will only be given a single
-  input parameter, the application instance. Your failure callbacks (on your
-  `catch` Promise chain) will only be given a single input parameter, the
-  error.
-
-  If you opt to mix and match the callback and the Promise interfaces, be
-  aware that all registered callbacks passed to the `ready` function will be
-  executed first, in the order in which they were attached, and then the
-  Promise chain will be resolved or rejected.
+  Note: If you opt to mix and match the callback and the Promise interfaces,
+  be aware that all registered callbacks passed *to* the `ready` function
+  will be executed first, in the order in which they were attached, and then
+  the Promise chain will be resolved or rejected, which will in turn trigger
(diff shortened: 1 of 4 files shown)


Velveeta:
Well, the design *is* intentional, and the reasoning was that if a client is supporting a mix of browsers, some of which may and some of which may not have Promise support, then they should be polyfilling their own environment to account for that, if they're planning to use Promise-style syntax. If they're not polyfilling, and are trying to use the Promise-style interface, and it breaks on some legacy browser they support, that's a problem that's technically on their end, because they're the ones consuming the code in a Promise fashion without polyfilling for it. They can just as easily use the `ready(callback)` syntax, and have it work in both Promise and non-Promise environments, but if they support non-Promise-compliant browsers, and are using the Promise-style interface without polyfilling for it on their end, then that's an implementation issue on their end.

savetheclocktower:
Sure. My only point — lodged not as a serious reservation, but just as something worth thinking about — is whether that will create support work for us down the road. The answer is probably unknowable.

Velveeta:
It's certainly a valid concern, and definitely isn't one that I know the answer to. The times we've polyfilled Promise ourselves are only when we're using our own Promises internally, not making Promise interfaces available externally for other people to consume. That being said, we had a pretty good discussion about this, and I personally feel pretty confident in this decision, but I'm going to ping @ceejtron, @tnunamak, and @aterranova-bv since they were the major contributors to that conversation yesterday, and see if they have any additional concerns around this in the context of potentially making more support work for ourselves later down the line by making this interface environment-dependent.

tnunamak:
I've been thinking about this and TBH, the pattern feels weird to me, but leaving polyfilling up to the user of bv-ui-core makes a lot of sense. I think we should at least make this very clear by bolding some of the documentation to call attention to it, or something.

lib/application/README.md
```diff
@@ -49,6 +69,17 @@ APP.exampleApp.render({ productId : 1 });
 
 ## Methods to Be Called by Core Application Code
 
+- `processReady(error)`: A method that your core application code can call in
+  order to access calls to `ready` that have occurred. The only parameter it
+  accepts is an optional one, for an error to be passed to the callbacks in
+  its queue, and to reject its Promise with, if used in an environment that
+  supports ES6 Promises. It will then replace the original definition of the
+  `ready` method. Queued items are processed synchronously. If your
+  application is being loaded via the BV Loader, the Loader will handle
```
tnunamak: Maybe a bit pedantic, but we shouldn't be creating references to BV Loader or other internal BV things in this repo, which is OSS.

lib/application/index.js
```diff
@@ -54,13 +56,38 @@ function processQueue (queue, fn, sync) {
...
 function Application (config) {
   this.config = config;
+  this._readyQueue = [];
   this._renderQueue = [];
   this._configQueue = [];
+  if (typeof global.Promise === 'function') {
+    this._promise = new Promise(function (resolve, reject) {
+      this._resolve = resolve;
+      this._reject = reject;
+    }.bind(this));
```
tnunamak: Is this a work-around to not having arrow functions?

Velveeta: Yep! :D

sirugh:
FYI the API module in this library also uses fetch and promises and expects the user of the module to polyfill as necessary.

Velveeta:
So, if that's an assumption that the library is making, that a consumer will have polyfilled for promises in order to use the api module, do we want to scrap the *check* for Promise support in this one and also just assume that promises are available, and provide the callback and promise interface as the standard?

sirugh:
>  do we want to scrap the check for Promise support in this one and also just assume that promises are available

Yep, and call it out in the readme for this module

https://github.com/bazaarvoice/bv-ui-core/tree/master/lib/api#requirements

tnunamak:
I suppose worst case, if anyone is currently using this module (Firebird?, Curations?) and not already polyfilling Promise, they can upgrade to these changes and either not use .ready() or add a polyfill.

PR #72, opened 2017-06-28
SWAT-1245 -- Safer Promise polyfill (ponyfill)

Velveeta:
[SWAT-1245](https://bits.bazaarvoice.com/jira/browse/SWAT-1245)

## Problem / Goal
Clients may sometimes polyfill Promise themselves with subpar implementations, which may cause our code that relies upon them to break.

## Solution
To account for cases where a client may have a preexisting version of a Promise polyfill that might not be spec-compliant, we're now polyfilling it ourselves if we detect anything other than native code. This also sneaks in some test code fixes, and explicitly imports our Promise polyfill wherever we're referencing Promise within the project. This should actually remove the need for polyfilling Promise in the scoutfile, which I've always hated that we had to do, since it's not using them directly.

## Verification
1. Load a browser that supports Promises natively (Chrome/FF/Safari).
2. Run the following in the JS console (this is the logic being tested against in the polyfill):
`typeof window.Promise === 'undefined' || !window.Promise.toString().match(/\[native code]/)`
3. Verify that you get a false output (this would result in the native Promise being returned).
4. Load a browser that doesn't support native Promises (IE 11).
5. Run the following in the JS console (this is the logic being tested against in the polyfill):
`typeof window.Promise === 'undefined' || !window.Promise.toString().match(/\[native code]/)`
6. Verify that you get a true output (this would result in the polyfilled Promise being returned).

## TODO
Write some actual test coverage around this. Since it's using a regex test for `[native code]`, we should be able to supply a stub function with that string in it, and verify that our polyfill either returns that function or another (the polyfilled version).


lib/api/statistics.js
```diff
@@ -3,6 +3,10 @@
  * @fileOverview Module for interacting with the statistcs.json endpoint.
  *
  */
+
+// Local
+const Promise = require('../polyfills/promise');
```
sirugh: If you are doing this can you modify the readme for this module to explain that `Promise` is polyfilled but `fetch` is not?

Velveeta:
@tnunamak @aterranova-bv @savetheclocktower @taylormck @danheberden ... I was talking to @sirugh about this, and would like to get some additional feedback about how we feel regarding polyfilling certain things within the bv-ui-core files themselves. E.g. we *provide* the polyfills for any projects that are consuming bv-ui-core, but the bv-ui-core files themselves didn't previously *use* them if they had need of something that might have needed polyfilling.

We had previously polyfilled scoutfile itself, because it relied on `bv-ui-core/lib/application`, which uses an internal Promise that needed polyfilling for IE11, rather than polyfilling it within the application file itself. That meant leaking implementation details from the application class into the scoutfile in order to know we needed to polyfill Promise there. I never liked the taste of that, because scoutfile itself doesn't use Promises directly.

The polyfill we provide here does environmental checking to ensure we're either using a native implementation, or our own provided polyfill, so we can guarantee using a good polyfill version (over something that might exist in the client environment and doesn't implement it to spec), and by consuming it wherever we're using Promises within the bv-ui-core files, it means we're handling our own patchwork, rather than having to polyfill at the consuming layer. I personally think it's a cleaner solution, but wanted to get some feedback about how we feel collectively.

lib/api/statistics.js
Velveeta: I'd like to actually ponyfill fetch within this repo also, and also in a non-polluting way, and then import that ponyfill wherever we're using fetch internally here... That would just be done on an upcoming ticket.

tnunamak:
* Isn't a reason that we didn't make lib/application code itself consume the polyfill because that could be forcing a consumer to eat those bytes even if they don't need or want them?
* [SWAT-1291](https://bits.bazaarvoice.com/jira/browse/SWAT-1291) should remove one abstraction layer between the polyfilling code and the polyfill-needing code.

lib/application/index.js
```diff
@@ -4,6 +4,9 @@
  * configuration and rendering.
  */
 
+// Local
+var Promise = require('../polyfills/promise');
```
tnunamak: Specifically, won't the inclusion of this line make our scout files bigger?

Velveeta: The scoutfile is polyfilling it already, itself... All this would do is shift those bytes from that project to this one.

sirugh: Yea. The real question here is whether we want bv-ui-core to automatically include the polyfill.

tnunamak: Idk. That seems pretty opinionated for a library that could be used by different consumers with different browser support levels.

Velveeta: That's true, but the extra bytes we're pushing, after minification and gzipping, only add 1131, which is pretty insignificant. I can see the point, though, and can remove these 2 import lines, in which case, no documentation will need to be updated.

tnunamak: 1kb is insignificant enough that I don't feel strongly about it. I defer to your judgement on this one 😺.

sirugh:
In my opinion we should be less opinionated.

Velveeta:
Ok, I removed the 2 lines that were importing the polyfill internally, that will be left up to the clients to do, based on their own browser support matrix.

lib/api/statistics.js
```diff
@@ -91,7 +91,7 @@ function getStatistics ({ productIds, environment, key, type, filters = {} }) {
   return Promise.all(
     productIdChunks.map(products => new Promise((resolve, reject) => {
       const requestUri = `${uri}&filter=ProductId:${products.join(',')}`
-      fetch(requestUri)
+      return fetch(requestUri)
```
sirugh: If inline_ratings updates bv-ui-core after this is merged they should make sure to test this change.

jaenriquez:
verified 👍

PR #73, opened 2017-08-01
SWAT-1609 -- Adding getOriginalConstructor function

Velveeta:
[SWAT-1609](https://bits.bazaarvoice.com/jira/browse/SWAT-1609)

## Problem/Goal
In some situations, clients are overriding prototype functions of primitives (or using Prototype.js, which does that), and some prototype functions we're relying on can throw unexpected errors.

## Solution
Added a method to fetch an original version of the constructor from a throwaway iframe.

## Verification
1. Check out this PR and run the tests.

### DIFF
--- a/lib/getOriginalConstructor/README.md
+++ b/lib/getOriginalConstructor/README.md
@@ -1,0 +1,13 @@
+# getOriginalConstructor
+
+This file exposes a utility function to get a reset version of a primitive
+constructor, in the event that a client has overridden or polyfilled prototype
+methods in their own primitives, or used a utility like Prototype.js, which
+does that on its own.
+
+## Usage
+```javascript
+var getOriginalConstructor = require('getOriginalConstructor');
+
+getOriginalConstructor(Array).prototype.forEach(someNodeList, callback);
+```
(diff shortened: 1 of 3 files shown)


lib/getOriginalConstructor/README.md
```diff
@@ -0,0 +1,13 @@
+# getOriginalConstructor
+
+This file exposes a utility function to get a reset version of a primitive
+constructor, in the event that a client has overridden or polyfilled prototype
```
sirugh: You may want to reword this. As an OS repo, "site" is probably more appropriate than "client".

test/unit/getOriginalConstructor/index.spec.js
```diff
@@ -0,0 +1,34 @@
...
+
+  Array.prototype.push = push;
+
+  var originalArray = getOriginalConstructor(Array);
+
+  it('Should return an unaltered constructor', function () {
+    // Verify that the original push function matches '[native code]'
+    expect(!!originalPush.toString().match(/\[native code]/)).to.equal(true);
+    // Verify that the overridden push function matches '[native code]'
+    expect(!!Array.prototype.push.toString().match(/\[native code]/)).to.equal(false);
```
sirugh: The comment here seems wrong.

Velveeta: Good catch, stupid copy/paste :D

PR #74, opened 2017-08-03
SWAT-1242 -- Adding fetch polyfill module

Velveeta:
[SWAT-1242](https://bits.bazaarvoice.com/jira/browse/SWAT-1242)

## Problem/Goal
We want to be able to use the `fetch` API in our products, but IE11 doesn't support it, and we don't want to unnecessarily polyfill if we natively support it. We also want to make sure that if an existing polyfill is in place, that we use our own official polyfill (in case the existing polyfill doesn't do it to spec). All of this should be done without interfering with the global namespace, so as not to disrupt any client code that may be dependent on their own fetch polyfill.

## Solution
Wrap whatwg-fetch in a module that exposes either the native fetch function (and supporting constructor classes for Headers/Request/Response), or the whatwg-fetch versions of fetch and its supporting constructors.

## Verification
1. Check out this PR.
2. Run the tests, which simulate both an environment with native support and an environment without native support.
3. If further verification is necessary, create a test script to load in your browser, which includes this polyfill file, and load in both Chrome and IE11. Chrome should export the native fetch function, and IE11 should export the polyfilled version.

### DIFF
--- a/lib/polyfills/README.md
+++ b/lib/polyfills/README.md
@@ -1,8 +1,8 @@
 # Polyfills!
 
 The polyfills contained in this folder can be imported as necessary. They should
-export either the built-in being polyfilled, if found on the global, or the
-polyfill.
+export either the builtin native function(s), if found on the global, or the
+polyfill itself.
 
 ## Usage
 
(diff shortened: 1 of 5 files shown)


KikoRamos:
QA Verified using RH 👍

PR #77, opened 2017-08-08
SWAT-1609 -- Bug fixes for IE11

Velveeta:
IE11 was throwing an error because the constructors hash was being keyed by constructor.name, which doesn't exist in IE11. I added a regex so that we could prefer constructor.name where it was supported, and fallback to regex'ing a name out of the function declaration as a backup for browsers where that wasn't supported. That fixes that issue.

Secondly, there was an issue with returning cached versions of these lookups. The original implementation would create an iframe, stick it in the head, return the constructor from it, and then remove it from the head. In future calls, the cached function would be returned, which would cause IE11 to throw an error about trying to execute code from freed scripts. The options to fix this were to create/attach/detach the iframe on every request to this function, or create the iframe on the first request, and leave it in the head permanently, which is the option I chose, as it's the more lightweight option, and shouldn't interfere with anything else on the page (we're already creating plenty of nodes for our own content anyway).

### DIFF
--- a/lib/getOriginalConstructor/index.js
+++ b/lib/getOriginalConstructor/index.js
@@ -8,6 +8,7 @@
 
 var constructors = {};
 var constructorNameRegExp = /function\s+([^\(\s]+)/
+var iframe;
 
 var getOriginalConstructor = function getOriginalConstructor (constructor) {
   var constructorName = constructor.name;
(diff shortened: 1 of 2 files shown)


sirugh:
I'd like to see a unit test for the case where `constructor.name` does not exist.

PR #80, opened 2017-09-29
SWAT-2071 -- Fixing a bug in getOriginalConstructor

Velveeta:
We were previously appending the iframe to document.head, but it seems that in some environments, that causes the iframe's `contentWindow` property to be null, and errors when trying to pull the primitive reference from it. We're now appending it to document.body instead, since it's an empty iframe anyway, with `src="about:blank"`, which appears to clear up that issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ module's directory.
 - [date.now](./lib/date.now)
 - [domainPolice](./lib/domainPolice)
 - [evented](./lib/evented)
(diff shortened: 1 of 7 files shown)


Velveeta:
I've verified it via the modified tests in this PR, as well as checking out the head of this branch locally for use in bv-loader (after refactoring that code to account for this new Promise interface) and verifying that everything passed in the loader tests, and that it loaded properly on the test pdp page, and confirming that it was using the Promise interface throughout that code from within devtools.

PR #81, opened 2017-10-11
Setting fetch up as the default export

Velveeta:
When polyfilled through webpack with the ProvidePlugin, it's importing the exported object as the definition of fetch, vs the the import's named `fetch` export. We're now exporting the `fetch` function itself as the default export, and attaching the other exports to its namespace for use as named exports, in case the consumer also needs access to Headers/Request/Response.

Updated tests to verify that it works as both a default import and a named import.


Velveeta:
This should be a patch-level version bump, btw. It's backwards compatible with the previous interface, and should be considered a bug fix to correct the expected default import behavior.

lib/polyfills/fetch.js
```diff
@@ -36,22 +36,62 @@ if (typeof global.fetch === 'undefined' || !global.fetch.toString().match(/\[nat
...
   // Export the polyfilled items that were housed in the global.self namespace.
-  module.exports = {
-    fetch: global.self.fetch,
-    Headers: global.self.Headers,
-    Request: global.self.Request,
-    Response: global.self.Response,
-  };
+  // fetch itself should be the default export, but we need to also make sure
+  // that we export the other items for destructuring in case they're needed.
+  Object.defineProperties(global.self.fetch, {
```
tnunamak: I think there's some potential to simplify a bit. Just messing around here, probably not the best approach:

```javascript
function constructProperties (parent) {
  const keys = ['fetch', 'Headers', 'Request', 'Response']

  keys.forEach(key => {
    properties[key] = {
      enumerable: true,
      writeable: false,
      value: parent[key]
    }
  })
 return properties
}

// ...
Object.defineProperties(global.self.fetch, constructProperties(global.self))
// ...
Object.defineProperties(global.self.fetch, constructProperties(global))
```

lib/polyfills/fetch.js
```diff
@@ -36,22 +36,62 @@ if (typeof global.fetch === 'undefined' || !global.fetch.toString().match(/\[nat
...
 }
 else {
   // fetch is natively supported, export it and its support constructors.
-  module.exports = {
-    fetch: global.fetch,
-    Headers: global.Headers,
-    Request: global.Request,
-    Response: global.Response,
-  };
+  // fetch itself should be the default export, but we need to also make sure
```
tnunamak: We're modifying the native fetch object? That doesn't seem safe.

Velveeta: Yes, we are modifying the window.fetch function, but for the properties we're adding, it should be fine, as that should be treated as a function by any regular consumer, and is only being treated as an object for the purposes of destructuring the module exports. The other option would be to make 4 individual modules in that directory, one to export each of fetch/Headers/Request/Response as the default export of itself, and then we'd have to polyfill those individually as needed.

tnunamak:
My [first comment](https://github.com/bazaarvoice/bv-ui-core/pull/81#discussion_r144064749) is now hidden, but it still stands as the `Object.defineProperties` calls haven't changed with the latest update.

tnunamak:
@Velveeta any QA verification recommendations?

tnunamak:
So now, if you want the fetch function, `require('fetch').fetch()` like before, or `require('fetch/fetch')()`, right?

Velveeta:
The Object.defineProperties calls are gone with the latest commit, make sure you're actually viewing the latest commit, [here's what the index file looks like now](https://github.com/Velveeta/bv-ui-core/blob/5a02847821585c4889dcde93a06691624de93ea4/lib/polyfills/fetch/index.js). As far as testing instructions, I added some tests to verify that the original structure tests all pass with the new (reverted) index file, and that each of the additional modules return a function. The best way to test is going to probably be to add this as a polyfill to a project that uses it (e.g. RH) and make sure it works, and then remove it as a webpack polyfill and add it as a module-level import and verify that it works.

And yes, if you want the fetch function itself, you can do one of the following:

```javascript
import fetch from 'bv-ui-core/lib/polyfills/fetch/fetch';
import { fetch } from 'bv-ui-core/lib/polyfills/fetch';
const fetch = require('bv-ui-core/lib/polyfills/fetch/fetch');
const fetch = require('bv-ui-core/lib/polyfills/fetch').fetch;
```

PR #83, opened 2017-12-06
SWAT-2616 -- Adding a couple more performance polyfills

Velveeta:
Adding `getEntries`, `getEntriesByType`, and `getEntriesByName`, along with test coverage.

### DIFF
--- a/lib/performance/getEntriesByName.js
+++ b/lib/performance/getEntriesByName.js
@@ -22,7 +22,7 @@ var isNativeSupported = (typeof nativeGetEntriesByName === 'function');
  *  Uses the native-supported performance.getEntriesByName function
  */
 function nativeImplementation (name, entryType) {
-  nativeGetEntriesByName.call(performance, name, entryType);
+  return nativeGetEntriesByName.call(performance, name, entryType);
 }
 
 /**
--- a/lib/performance/getEntriesByName.js
+++ b/lib/performance/getEntriesByName.js
@@ -1,0 +1,60 @@
+/**
+ *  @fileOverview
+ *  Performance.getEntriesByName uses the native User Timing API
+ *  when available and falls back to a manual implementation if not
+ */
+
+// Imports
+var global = require('../global');
+var perfGetEntries = require('./getEntries').getEntries;
+
+// Determine now (once) if we can use the native User Timing API
+
+// Cache references to native implementation in case they are modified later
+var performance = global.performance;
+var nativeGetEntriesByName = performance && performance.getEntriesByName;
+
+// Is the User Timing API natively supported?
+var isNativeSupported = (typeof nativeGetEntriesByName === 'function');
+
+/**
+ *  Native Implementation
+ *  Uses the native-supported performance.getEntriesByName function
+ */
+function nativeImplementation (name, entryType) {
+  nativeGetEntriesByName.call(performance, name, entryType);
+}
+
+/**
+ *  Polyfill Implementation
+ *  Manually provides similar functionality to native
+ *  performance.getEntriesByName using a filter operation on
+ *  a list of all performance marks
+ */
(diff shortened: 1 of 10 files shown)


test/unit/performance/index.spec.js
```diff
@@ -8,6 +8,14 @@ var perfModule = require('../../../lib/performance');
 
 describe('lib/performance', function () {
 
+  it('exposes a getEntries function', function () {
+    expect(perfModule.getEntries).to.be.a('function');
+  });
+
+  it('exposes a getEntriesByName function', function () {
```
tnunamak: Should we have one for getEntriesByType too?

lib/performance/mark.js
```diff
@@ -7,6 +7,8 @@
 // Imports
 var global = require('../global');
 var perfNow = require('./now.js');
+var marks = require('./entries').marks;
```
tnunamak: This property is currently not read by any other module, correct?

lib/performance/getEntriesByName.js
```diff
@@ -0,0 +1,60 @@
...
+ *  Manually provides similar functionality to native
+ *  performance.getEntriesByName using a filter operation on
+ *  a list of all performance marks
+ */
+function polyfillImplementation (name, entryType) {
+  if (arguments.length === 0) {
+    throw new TypeError('Failed to execute \'getEntriesByName\' on \'Performance\': 1 argument required, but only 0 present.');
+  }
+
+  return perfGetEntries().filter(function (entry) {
```
tnunamak: Is filter guaranteed to exist?

lib/performance/getEntriesByName.js
```diff
@@ -0,0 +1,60 @@
...
+}
+
+module.exports = {
+  /**
+   *  @param {String} name - The name of the mark to filter for
+   *  @param {String} [entryType] - The entryType of the mark to filter for
+   */
+  getEntriesByName: function (name, entryType) {
+    // Note: because the implementation functions explicitly check their number of arguments,
+    // we want to preserve the calls exactly as they came in
```
tnunamak: What if someone calls with name but omits entryType? Does that need to be handled explicitly too?

lib/performance/getEntriesByName.js
Velveeta: In all of our supported browsers, yes... It's IE9+.

lib/performance/getEntriesByName.js
Velveeta: `entryType` is not a required parameter. If they call `getEntriesByName` with only a `name` parameter, and there are multiple types with the same name (possible, but unlikely), they would all come back in an array.

lib/performance/mark.js
Velveeta: Looking through the code, that does appear to be the case. I think once the timeline started being used in multiple modules, it made sense to me to also house the marks object separately in case of the same needs, but I'm not even importing it as part of the test modules. Also, I'm not even sure that `marks` is even needed at all anymore, because we never previously exposed a method to `getEntries`. It looks like any marks that were set with the polyfill method were simply inaccessible. The new getEntries* family of methods all use the timeline object, which we're also pushing mark records into within the mark module, so that object may be obsolete at this point.

test/unit/performance/index.spec.js
Velveeta: Yes we should, I'll get that added to this PR.

lib/performance/getEntriesByName.js
tnunamak: Oops, I misread the code. Looks good.

Velveeta:
For now, I'd say if it ain't broke, don't fix it. I'd rather not remove marks and risk having something break unexpectedly, and it only takes up a few extra bytes. I may revisit this in the future, and remove it at that time.

PR #85, opened 2018-01-29
Add a performance.measure polyfill

Velveeta:
Adding a polyfill for performance.measure, as well as removing some cruft that was [pointed out in the previous PR](https://github.com/bazaarvoice/bv-ui-core/pull/83#discussion_r160748671), and adding test coverage.

### DIFF
--- a/lib/performance/measure.js
+++ b/lib/performance/measure.js
@@ -1,0 +1,94 @@
+/**
+ *  @fileOverview
+ *  Performance.measure uses the native User Timing API
+ *  when available and falls back to a manual implementation if not
+ */
+
+// Imports
+var global = require('../global');
+var perfNow = require('./now.js').now;
+var perfGetEntriesByType = require('./getEntriesByType').getEntriesByType;
+var timeline = require('./entries').timeline;
+
+// Determine now (once) if we can use the native User Timing API
+
+// Cache references to native implementation in case they are modified later
+var performance = global.performance;
+var nativeMeasure = performance && performance.measure;
+
+// Is the User Timing API natively supported?
+var isNativeSupported = (typeof nativeMeasure === 'function');
+
+/**
+ *  Native Implementation
+ *  Uses the native-supported performance.measure function
+ */
+function nativeImplementation (name, startMark, endMark) {
+  return nativeMeasure.call(performance, name, startMark, endMark);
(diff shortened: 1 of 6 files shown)


lib/performance/measure.js
```diff
@@ -0,0 +1,94 @@
...
+function polyfillImplementation (name, startMark, endMark) {
+  // Validation: a measure name must be provided
+  if (arguments.length < 1) {
+    throw new TypeError('Failed to execute \'measure\' on \'Performance\': 1 argument required, but only 0 present.');
+  }
+
+  var marks = perfGetEntriesByType('mark');
+  var filteredMarks;
+
+  // if no startMark was passed, used
```
tnunamak: Looks like some of this comment was lef

Velveeta: ahhhhh ye

lib/performance/measure.js
```diff
@@ -0,0 +1,94 @@
...
+  var startTime = 0;
+  var endTime = perfNow();
+
+  if (startMark !== undefined) {
+    filteredMarks = marks.filter(function (entry) {
+      return entry.entryName === startMark;
+    });
+    if (!filteredMarks.length) {
+      throw new Error('Failed to execute \'measure\' on \'Performance\': The mark \'foosdfsdf\' does not exist.');
```
tnunamak: foosdfsdf

lib/performance/measure.js
```diff
@@ -0,0 +1,94 @@
...
+    // use the most recent mark with the matching name
+    startTime = filteredMarks[filteredMarks.length-1].startTime;
+  }
+
+  if (endMark !== undefined) {
+    filteredMarks = marks.filter(function (entry) {
+      return entry.entryName === endMark;
+    });
+    if (!filteredMarks.length) {
+      throw new Error('Failed to execute \'measure\' on \'Performance\': The mark \'foosdfsdf\' does not exist.');
```
tnunamak: More foosdfsdf

lib/performance/measure.js
Velveeta: Ha! whoops :D

Velveeta:
Ok, pushed changes to fix those items, good catches :)

PR #88, opened 2018-09-06
CCS-29198: Fixing URI parsing regex for @ symbol.

Noirbot:
The short version of the issue here is that both regexes were looking for the `@` character as a delimiter for a userInfo string as part of the URI's authority section. Something like:

```
https://bob:smith@subdomain.example.com
```

This is totally valid as a URL, where it should parse `userInfo` as `bob:smith`. The issue is that when it saw a URL like:

```
https://subdomain.example.com:8080/?user=tim@bv.com
```

It would parse out the `userInfo` as `subdomain.example.com:8080/?user=tim` and then try to parse the rest of the uri from `bv.com`, which is clearly wrong.

As far as I can tell, the way around this is to rely on the fact that an unescaped `/` character cannot come in between the `protocol` segment, e.g. `http://` and the `@` that could delimit the userInfo segment. So a URI like 

```
https://bob:sm/ith@subdomain.example.com
```

would not be valid. If your username or password had a `/` in it, you would need to use the proper escape code instead of the `/` literal. 

As far as I can tell, this serves to fix the issues we were seeing with the Jack Wills URIs.

## HOLY RISK BATMAN

Ok, there are so many ways this is a terrifying code change to merge. Just off the top, I'm hand-editing a regular expression, written by someone on the internet, with 13 capture groups, that's over 100 characters long, and meant to reflect a 61-page RFC. This Regex, when it messes up, will prevent all BV content from rendering on a page. If I've introduced a bug here, it'll black out entire sites, but potentially in ways that are pretty hard to find and notice. In addition, we've clearly seen that the base regex here is fallible, since we're in this position in the first place.

I'm going to see about adding a TON more unit tests for this, based on https://rodneyrehm.de/t/url-regex.html We'll see if that finds anything. Otherwise, we may need to do a pretty robust real-world pass of testing on this.


danheberden:
So I get that we want this to be a general utility, but it seems we might be doing a lot of work that the browser already does for us.

[Location](https://developer.mozilla.org/en-US/docs/Web/API/Location) gives us `hostname`, which gives us something easier to match against.

danheberden:
I imagine just in the spirit of being general purpose. Sounds like this change, as it's formatted, is the right one. It might make sense to follow up with a PR for 

```
function parseUri (str, locationOnly=false) {}
```

type thing so we can pass only location (or just pass a `window.location` object.

msmolev:
Silly question, why not use URL object for parsing? 
https://nodejs.org/api/url.html#url_url

danheberden:
/me checks IE11 stats hoping they're abysmal (they won't be) :p

msmolev:
For IE11 we could create ```a``` element and set href to whatever we're parsing 
```var checkURL = document.createElement("a");
    checkURL.href = href;
checkURL.hostname```

**Comment by danheberden** [comment]
@Noirbot would it help to get a list of high-traffic urls to test against?

**Comment by danheberden** [comment]
@msmolev what sort of black magic…

File: lib/parseUri/index.js [hunk 1]
```diff
@@ -31,9 +31,13 @@ parseUri.options = {
...
     parser: /(?:^|&)([^&=]*)=?([^&]*)/g
   },
+  // Both of these parser regexes have been edited from the original linked above.
+  // A / character has been added to the excluded characters in the 4th and 5th
+  // capture groups to fix a bug where it would see a @ character in a query
+  // and consume it like it was delimiting a userInfo section.
   parser: {
-    strict: /^(?:([^:\/?#]+):)?(?:\/\/((?:(([^:@]*)(?::([^:@]*))?)?@)?([^:\/?#]*)(?::(\d*))?))?((((?:[^?#\/]*\/)*)([^?#]*))(?:\?([^#]*))?(?:#(.*))?)/,
-    loose: /^(?:(?![^:@]+:[^:@\/]*@)([^:\/?#.]+):)?(?:\/\/)?((?:(([^:@]*)(?::([^:@]*))?)?@)?([^:\/?#]*)(?::(\d*))?)(((\/(?:[^?#](?![^?#\/]*\.[^?#\/.]+(?:[?#]|$)))*\/?)?([^?#\/]*))(?:\?([^#]*))?(?:#(.*))?)/
+    strict: /^(?:([^:\/?#]+):)?(?:\/\/((?:(([^\/:@]*)(?::([^\/:@]*))?)?@)?([^:\/?#]*)(?::(\d*))?))?((((?:[^?#\/]*\/)*)([^?#]*))(?:\?([^#]*))?(?:#(.*))?)/,
```
robbytx: Seems like a reasonable fix to me.  

For consistency with the other character classes, I suggest using ``[^:\/@]`` instead of ``[^\/:@]``.  

Also, should we not include `?` and `#` as excluded characters?  Is ``https://www.example.com?user=a@b.com`` a valid URL??

test/unit/parseUri/index.spec.js
```diff
@@ -50,4 +50,25 @@ describe('lib/parseUri', function () {
     expect(parsedUri.queryKey).to.eql({ foo: 'bar', baz: '3' });
   });
 
+  it('correctly parses a URL with the @ symbol in the query', function () {
+    var url = 'https://dev04-webstore-jackwills.demandware.net/on/demandware.store/Sites-jwuk-Site/en_GB/DebugPipelines-ShowConfirmation?cid=p@p.com&&oid=00026203';
```
robbytx: We should not promote use URLs that belong to other organizations.  Please replace with an example URL that demonstrates the problem.

lib/parseUri/index.js
Noirbot: I can definitely do the consistency fix.

From my experimentation, it seems like `https://www.example.com?user=a@b.com` is not valid. Browsers auto-correct it to `https://www.example.com/?user=a@b.com`, which would parse correctly here.

robbytx: 👍 I also confirmed this in IE 11.

PR #90, opened 2019-01-03
CCS-31771: Removing ParseUri and Fixing DomainPolice.

Noirbot:
## The Problem

Parsing URIs is a fraught thing to try to do. For years, we got by using a regular expression, but over the last few months, we've seen more and more edge cases with it that are causing issues for real clients. There's no easy way to fix this, since the actual spec for URIs isn't regular. To reliably achieve what parseUri wants to do, you need a much more robust library, most of which are too large to be used in the client-side contexts that consumers like Domain Police expect.

## The Solution

The most reliable way to parse out a URI, when you're running in a browser, is to just leverage the [Location interface](https://developer.mozilla.org/en-US/docs/Web/API/Location), which provides `window.location.hostname` which is exactly what we were using parseUri to get out for us.

In general, this provides a better solution by pushing the responsibility to parse the URI onto the user of Domain Police. In a browser context, we can use the Location interface and reduce our import size. In a Node.js environment, you can import a robust library to use, since you're less worried about bundle sizing.

## Notes

I've also taken this opportunity to clean up this repo some, both in terms of documentation, and ownership, to reflect the current state of the project with more honesty.


taylormck:
@Velveeta ❤️

sirugh:
Am I too late for the party? Darn 😭

PR #95, opened 2019-09-25
apply CSP fix to v0.5.0 version

bodrovphone:
**Problem**

We migrated scoutfile dependency 2.0.0 -> 4.0.1  on Firebird repo to apply CSP-compatible fix.

old scoutfile(v2.0.0) has bv-ui-core v0.5.0 as its dependency
new  scoutfile(v4.0.1) includes bv-ui-core v2.1.1

The problem here with [this changes](https://github.com/bazaarvoice/bv-ui-core/commit/8ac7f0a0a470276cdc3800a80553cf7a1a971ca4) on bv-ui-core 0.11.1 which breaks Site Preview and mixed integration (Firebird + SWAT).

**Solution**

A solution would be to apply this changes on bv-ui-core v0.5.0 and release bv-ui-core v0.5.1 containing these changes.

Based on this new package we can update scoutfile and that should be enough :)

### DIFF
--- a/lib/global/index.js
+++ b/lib/global/index.js
@@ -1,11 +1,16 @@
 /**
  *  @fileOverview Provides a reference to the global object
- *
- *  Functions created via the Function constructor in strict mode are sloppy
- *  unless the function body contains a strict mode pragma. This is a reliable
- *  way to obtain a reference to the global object in any ES3+ environment.
- *  see http://stackoverflow.com/a/3277192/46867
+ *  the below solution works in ES3+ environment and doesn't violates CSP in Chrome apps 
+ *  see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis
  */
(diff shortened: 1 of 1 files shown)


bodrovphone:
Hi @tnunamak,

Can you please review and merge it with 0.5.1 branch?

I made these steps just in case:

1) Create branch 0.5.1
`cd bv-ui-core`
`git checkout -b 0.5.1 524965d8c6bbc1a6e066beaf47b89c4c92bfb8fa`
`git push bv --set-upstream 0.5.1`
2) then merge this PR with 0.5.1 you created with steps above
3) `git pull bv 0.5.1` to pick up merged changes locally
4) after that you'll need to publish v0.5.1 from 0.5.1 branch
I believe you can use it like so: 
`npm version 0.5.1`
`npm publish --tag firebird-csp`   
tag above is needed to not make this package as latest.

Please let me know if I missed anything.

Thanks a lot!

PR #97, opened 2019-11-01
Prefer window over self as the global object

tnunamak:
Safari 12.0 does not support `globalThis`, so before this change, `self` was selected as the global in that environment. I observed a case where the global context was then defined to be a `<script>` object but where the window would be preferable.

Any objections?

Btw, [this article](https://mathiasbynens.be/notes/globalthis) describes some of the differences of the different values used here.

### DIFF
--- a/lib/global/index.js
+++ b/lib/global/index.js
@@ -7,8 +7,8 @@
 
 var getGlobal = function () {
   if (typeof globalThis !== 'undefined') { return globalThis; }
-  if (typeof self !== 'undefined') { return self; }
   if (typeof window !== 'undefined') { return window; }
+  if (typeof self !== 'undefined') { return self; }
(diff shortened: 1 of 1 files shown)


haileybee1231:
As long as we can be sure that this doesn't have any knock-on affects in other browsers, I'm fine signing off on this.

@tnunamak have you verified in other environments? I imagine we should be fine because most of them can just get `globalThis`, but just in case...

tnunamak:
I haven't verified this, no.

robbytx:
> I observed a case where the global context was then defined to be a <script> object but where the window would be preferable.

I wonder if this was due to a "bug" in submission container where we were leaking a ``self`` variable into the global namespace that pointed at the current ``<script>`` tag.

PR #99, opened 2020-03-26
SWAT-5709: Revise default and custom error handler behavior

robbytx:
[SWAT-5709](https://bits.bazaarvoice.com/jira/browse/SWAT-5709) https://github.com/bazaarvoice/bv-ui-core/issues/96

## Problem

The previous pull request for this issue (#98) had a number of problems, which I've logged.

## Solution

Address each of the problems by simplifying the error handling process.

## Verification

I've updated the unit test to cover most of the new behavior, and I've also verified the new behavior locally by adding code to the BV Loader.

### DIFF
--- a/lib/evented/index.js
+++ b/lib/evented/index.js
@@ -20,36 +20,23 @@ var nativeIndexOf = ArrayProto.indexOf;
 var slice = ArrayProto.slice;
 
 /**
- * Initialize a custom error handler or default to the BV Loader error reporting
- * regardless of application. This will automatically be called at the first
- * invocation of the ".on" method, and can be 
- * event
+ * By default, we should simply log the error to the console.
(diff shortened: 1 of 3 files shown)


lib/evented/index.js
```diff
@@ -114,9 +96,9 @@ function trigger(event) {
       events[event][i].apply(this, args);
     }
     catch (e) {
-      this.errorHandler(e, {
+      (this.eventErrorHandler || defaultErrorHandler).call(this, e, {
         event: event,
-        details: arguments
+        data: args
```
robbytx: I've renamed this for consistency with the documented name on the README.

robbytx:
Since this is adding new functionality, it should be released as 2.2.0

PR #102, opened 2020-06-12
CCS-36556: added config loaderOptions

vmukhin17:
Added config loaderOptions for staticAssetLoader

### DIFF
--- a/lib/staticAssetLoader/index.js
+++ b/lib/staticAssetLoader/index.js
@@ -92,7 +92,7 @@ module.exports = {
 
     // TODO: think about allowing some of the loader options to be set via the
     // static asset loader config object.
-    var loaderOptions = {};
+    var loaderOptions = config.loaderOptions || {};
     var NS = namespacer.namespace(namespaceName);
     var registry = NS._staticAssetRegistry = NS._staticAssetRegistry || {
       // Stores defined entities by asset name. E.g.:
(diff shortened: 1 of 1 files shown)


lib/staticAssetLoader/index.js
```diff
@@ -92,7 +92,7 @@ module.exports = {
 
     // TODO: think about allowing some of the loader options to be set via the
```
robbytx, line 93: Seems like we can remove this comment?

robbytx:
@vmukhin17 can you give an example of the configuration option(s) that you expect to pass?

vmukhin17:
We need to increase default timeout from 10000 to 30000 -https://github.com/bazaarvoice/firebird/pull/6478/files#diff-2cf140d836ad55960d0acabc32583229R36

robbytx:
Apparently, the [2.3.0 release commit](https://github.com/bazaarvoice/bv-ui-core/tree/v2.3.0) was never pushed into master. 🤷‍♂️

So I've rebased the commit from this PR (30cfebd) onto that one, and force-pushed master, making the commit for this PR into: https://github.com/bazaarvoice/bv-ui-core/commit/74cf5ceda366d403640ba4281c079934e487ecc4

robbytx:
Ok... this change was released as a minor version, since it brings new functionality.  It's now released and available: https://www.npmjs.com/package/bv-ui-core/v/2.4.0

PR #110, opened 2021-08-24
[SWAT-7003] Cookie and Cookie consent module fixes

soumen-saniel:
## JIRA Ticket
[SWAT-7003](https://atl-reference.bazaarvoice.io/jira/browse/SWAT-7003)

## Requirement
- Ability to switch off cookie consent.
- Add 'change' event to valid events for cookie consent.
- Remove fallback local store for unset cookies.

## Solution
- Consent will be off by default so that if cookie module is being used without the consent being initialized, it should work as expected.
- Added one more parameter to initConsent method to pass a boolean that can enable/disable consent.
- Added 'change' event to valid events for cookie consent.
- Removed fallback local store for unset cookies from the cookie module.

### DIFF
--- a/lib/cookie/README.md
+++ b/lib/cookie/README.md
@@ -10,7 +10,7 @@ console.log(cookie.read('RememberMe')); // '1'
 cookie.remove('RememberMe');
 ```
 
-The following methods are provided:
+## The following methods are provided:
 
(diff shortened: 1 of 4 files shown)


soumyadipsarbajna (approved):
LG

PR #120, opened 2022-06-22
Adding Device checks

anishbvoice:
This PR is for Adding device check in to the library.
These two functions are repeatedly getting written in all the swat applications. 
We can keep it in the same place. 

Usage: 
var Device = require('bv-ui-core/lib/cookie');
Device.isMobile();
Device.isTablet();

### DIFF
--- a/lib/checkDevice/README.md
+++ b/lib/checkDevice/README.md
@@ -1,0 +1,10 @@
+#checkDevice
+This is a small method to check whether the device user is using Mobile/Tablet. 
+
+
+##Usage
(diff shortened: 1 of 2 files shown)


lib/checkDevice/README.md
```diff
@@ -0,0 +1,10 @@
+#checkDevice
+This is a small method to check whether the device user is using Mobile/Tablet.
```
anishbvoice, line 2: Please suggest better lines. 
Thanks in advance

PR #126, opened 2023-03-29
Adding messageFormat module for INTL API 

RahulSubramaniam17:
# PR description

## JIRA tickets

[PD-191705](https://bazaarvoice.atlassian.net/browse/PD-191705)

## This PR is for

- Feature 

## Requirements / issues

- We have a functionality to convert locale to a country name. 
- For this we are using INTL.DisplayNames. Since this function will be used in multiple places and it is using the browser API . We can have the code in bv-ui-core

## Solutions

- Using the INTL.DisplayNames API . We can convert the locale  to a country. 
-

## Unit tests

- Added messageFormat unit test cases 
- Added Chrome Headless as a browser to run unit test cases instead of PhantomJS as INTL was not available in PhantomJS
- Fixed Test cases for performance module 

## Steps to verify

- Copy paste the function in the browser and check if it is performing as expected 
- 
Screenshot 2023-03-29 at 12 01 06 PM

## Check points

- [x] [Design / approach [doc](https://bazaarvoice.atlassian.net/wiki/spaces/DEV/pages/78216462383/CFR+21+-+Display+Native+reviews+Design+Doc)
- [x] Functionality

- [x] Unit test cases

## Version 
This will be released as a new minor version as we are adding a new module 

[PD-191705]: https://bazaarvoice.atlassian.net/browse/PD-191705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/messageFormat/index.js
+++ b/lib/messageFormat/index.js
@@ -20,16 +20,21 @@ function generateCountryFromLocale (locale) {
     return '';
   }
     
-    // Split the locale code into language and region components
-  var localeArr = locale.split('_');
-  if (localeArr.length < 2) {
+  // Split the locale code into language and region components
+  try {
+    var localeArr = locale.split('_');
+    if (localeArr.length < 2) {
+      return '';
+    }
+  // Get the display name for the region (i.e., the country name)
+    var regionNames = new global.Intl.DisplayNames(localeArr[0], { type: 'region' });
+    var countryName = regionNames.of(localeArr[1]);
+  }
+  catch (error) {
     return '';
   }
-    // Get the display name for the region (i.e., the country name)
-  var regionNames = new global.Intl.DisplayNames(localeArr[0], { type: 'region' });
-  var countryName = regionNames.of(localeArr[1]);
-    // Return the country name, or an empty string if the region is not valid
-  return countryName || '';
+  // Return the country name, or an empty string if the region is not valid
+  return countryName 
 }
 
 module.exports = {
--- a/lib/messageFormat/index.js
+++ b/lib/messageFormat/index.js
@@ -9,7 +9,8 @@
  */
 
 
-//var global = require('../global');
+var global = require('../global');
+
 /**
  * Generates the country name associated with the given locale
  * @param {string} locale The locale code (e.g., 'en_US')
--- a/lib/messageFormat/index.js
+++ b/lib/messageFormat/index.js
@@ -9,34 +9,47 @@
  */
 
 
+//var global = require('../global');
 /**
  * Generates the country name associated with the given locale
- * @param {string} locale The locale code (e.g., 'en-US')
+ * @param {string} locale The locale code (e.g., 'en_US')
  * @returns {string} The name of the country associated with the locale, or an empty string if the locale is not supported by Intl.DisplayNames
  */
-export function generateCountryFromLocale(locale) {
-
-    //PolyFill for Intl.DisplayNames
-    if (typeof Intl === 'undefined' || typeof Intl.DisplayNames === 'undefined') {
-      // Load the polyfill if Intl or Intl.DisplayNames is not defined
-      const polyfillUrl = 'https://cdn.polyfill.io/v3/polyfill.min.js?features=Intl.DisplayNames';
-      const scriptEl = document.createElement('script');
-      scriptEl.src = polyfillUrl;
-      scriptEl.async = false;
-      document.head.appendChild(scriptEl);
-    }
-    // // Check if Intl.DisplayNames is supported
-    // if (!Intl.DisplayNames) {
-    //   return '';
+function generateCountryFromLocale (locale) {
+  if (!locale || locale.length === 0) {
+    return '';
+  }
+  
+    // Currently commenting this code , remove if polyfill is not required
+  
+    // // PolyFill for Intl.DisplayNames
+    // if (
+    //   typeof global.Intl === "undefined" ||
+    //   typeof global.Intl.DisplayNames === "undefined"
+    // ) {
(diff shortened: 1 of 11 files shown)


lib/messageFormat/index.js
```diff
@@ -0,0 +1,58 @@
...
+ *  messageFormat provides utility functions to which use the INTL API which provides 
+ *  language sensitive string comparison, number formatting, and date and time formatting.
+ * 
+ *  messageFormat can have functions which are used in multiple places
+ *
+ *  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl
+ */
+
+
+var global = require('../global');
```
soumen-saniel, line 12: Use `const` and `let` instead of `var` throughout the file
We don't need to follow the previously followed conventions unless the build fails

lib/messageFormat/index.js
```diff
@@ -0,0 +1,58 @@
...
+    //     "https://cdn.polyfill.io/v3/polyfill.min.js?features=Intl.DisplayNames";
+    //   const scriptEl = document.createElement("script");
+    //   scriptEl.src = polyfillUrl;
+    //   scriptEl.async = false;
+    //   scriptEl.onerror = () => {
+    //     console.log("Failed to Load Intl.DisplayNames Polyfill");
+    //     polyfillLoaded = false; // set flag to false if the polyfill fails to load
+    //   };
+    //   document.head.appendChild(scriptEl);
+    // }
```
soumen-saniel, line 42: I don't think we need polyfill as Intl.DisplayNames has support in all major browsers https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames

test/unit/performance/getEntriesByType.spec.js
```diff
@@ -9,13 +9,15 @@ var perfMark = require('../../../lib/performance/mark.js');
...
+    //Clearing all the marks as marks from other tests were being added
+    performance.clearMarks()
     perfMark.mark('test-by-name1');
     perfMark.mark('test-by-name2');
   });
 
   it('returns an array of all matching performance marks by type', function () {
     var result = perfGetEntriesByType.getEntriesByType('mark');
-
+    console.log(result)
```
soumen-saniel, line 20: Remove the console statement

soumen-saniel (changes requested):

lib/messageFormat/index.js
RahulSubramaniam17: The transpiler does not support let and const

lib/messageFormat/index.js
```diff
@@ -0,0 +1,37 @@
...
+ * Generates the country name associated with the given locale
+ * @param {string} locale The locale code (e.g., 'en_US')
+ * @returns {string} The name of the country associated with the locale, or an empty string if the locale is not supported by Intl.DisplayNames
+ */
+function generateCountryFromLocale (locale) {
+  if (!locale || locale.length === 0) {
+    return '';
+  }
+    
+    // Split the locale code into language and region components
```
akshay-choulwar-94, line 23: Fix the indenttation of the comments.

lib/messageFormat/index.js
```diff
@@ -0,0 +1,37 @@
...
+  var localeArr = locale.split('_');
+  if (localeArr.length < 2) {
+    return '';
+  }
+    // Get the display name for the region (i.e., the country name)
+  var regionNames = new global.Intl.DisplayNames(localeArr[0], { type: 'region' });
+  var countryName = regionNames.of(localeArr[1]);
```
akshay-choulwar-94, line 30: This line will throw the error if you pass the invalid locale string like `aa_bb`

lib/messageFormat/index.js
akshay-choulwar-94: @RahulSubramaniam17 We don't have support of es6? If yes then we need to think about to upgrade it may be in tech debt @soumen-saniel

lib/performance/getEntriesByName.js
```diff
@@ -53,7 +53,8 @@ module.exports = {
   getEntriesByName: function (name, entryType) {
     // Note: because the implementation functions explicitly check their number of arguments,
     // we want to preserve the calls exactly as they came in
-    var hasArguments = (arguments.length > 0);
+   
+    var hasArguments = (Object.values(arguments).length > 0);
```
akshay-choulwar-94, line 57: Why this change?

package.json
```diff
@@ -52,6 +52,9 @@
     }
   },
   "dependencies": {
```
akshay-choulwar-94, line 54: Why these packages?

package.json
```diff
@@ -12,7 +12,7 @@
   },
   "scripts": {
     "lint": "eslint karma.conf.js lib test",
-    "test": "./node_modules/karma/bin/karma start --single-run --browsers PhantomJS",
+    "test": "./node_modules/karma/bin/karma start --single-run --browsers ChromeHeadless",
```
akshay-choulwar-94, line 15: Only for test cases we are using chromeheadless browser in this task only. My thought is like why we can't create the separate task to update the phantomJS to ChromeHeadless. Atleast to track it we should have ticket created for it. @soumen-saniel your thoughts

soumen-saniel (approved):
Looks good

Shubhakar-A (approved):
Looks good to me

akshay-choulwar-94 (approved):
Looks good to me

PR #133, opened 2023-08-04
[utility fix] fix issue when global is polluted

Shubhakar-A:
# PR description

## JIRA tickets

[PD-215136](https://bazaarvoice.atlassian.net/browse/PD-215136)
[Approach document](https://bazaarvoice.atlassian.net/wiki/spaces/DEV/pages/edit-v2/78462714460)

## This PR is for

- Bug

## Requirements / issues

- When a client uses transpilations for its esModule files with default exports, these transpilations sets __esModule to true to each of such module. For Example: If the client has a esModule which exports globalThis/global/window object by default. It sets __esModule to true, to value after resolution of the modulie, in this case, window.__esModule is set
- SWAT apps uses webpack as bundler, webpack adds logic to import such default export esModule's by a condition, if esModule is set, then return module's default.
- This causes issue as this tries to return windows.default , which do not exist, when importing global.js 

## Solutions

- Added a straightforward check for the same global object for __esModule, if set add create proxy object to whatever is returned
- Add getters and setter for this object.
- For getter, add a special case to check default, if requested, return the same object

## Unit tests

- Global object
  - should export window with default pointing to windows if __esModule is set

## Version

- Patch - 2.8.1

[PD-215136]: https://bazaarvoice.atlassian.net/browse/PD-215136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/global/index.js
+++ b/lib/global/index.js
@@ -19,7 +19,21 @@ var getGlobal = function () {
     global import in all bundle use cases
   */
   if (globalObj && globalObj.__esModule) {
-    globalObj.default = Object.assign({}, globalObj)
+    const override = {
+      get: function (target, prop) {
+        if (prop === 'default') {
+          return target
+        }
+
+        return target[prop]
+      },
+
+      set: function (target, prop, value) {
+        target[prop] = value
+      },
+    };
+    const proxyGlobal = new Proxy(globalObj, override)
+    return proxyGlobal
   }
   return globalObj
 };
--- a/lib/global/index.js
+++ b/lib/global/index.js
@@ -20,16 +20,16 @@ var getGlobal = function () {
   */
   if (globalObj && globalObj.__esModule) {
     const override = {
-      get: function (target, prop) {
+      get: function (target, prop, receiver) {
         if (prop === 'default') {
           return target
         }
 
-        return target[prop]
+        return Reflect.get(receiver, prop, target)
       },
 
       set: function (target, prop, value) {
-        target[prop] = value
+        Reflect.set(target, prop, value)
       },
     };
     const proxyGlobal = new Proxy(globalObj, override)
--- a/lib/global/index.js
+++ b/lib/global/index.js
@@ -6,11 +6,22 @@
 'use strict'; /* eslint strict:0 */
 
 var getGlobal = function () {
-  if (typeof globalThis !== 'undefined') { return globalThis; }
-  if (typeof window !== 'undefined') { return window; }
-  if (typeof self !== 'undefined') { return self; }
-  if (typeof global !== 'undefined') { return global; }
-  throw new Error('unable to locate global object');
+  var globalObj = (function () {
+    if (typeof globalThis !== 'undefined') { return globalThis; }
+    if (typeof window !== 'undefined') { return window; }
+    if (typeof self !== 'undefined') { return self; }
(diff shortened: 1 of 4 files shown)


RahulSubramaniam17 (commented):
There are many cases where the BV object is being used without accessing global in swat reviews. We should raise a PR to change that was well 
Screenshot 2023-08-04 at 11 19 33 AM

Shubhakar-A:
> There are many cases where the BV object is being used without accessing global in swat reviews. Should we raise a PR to change that was well Screenshot 2023-08-04 at 11 19 33 AM

Agreed. We can make these changes to keep it consistent, however, this will not break as now we expose windows object as is and windows.default = windows.

akshay-choulwar-94 (approved):
Looks good to me.

anishbvoice (approved):
Looks good to me

RahulSubramaniam17 (approved):
LGTM

vignesh-ph (approved):
Looks good to me

lib/global/index.js
```diff
@@ -6,11 +6,22 @@
...
+    if (typeof global !== 'undefined') { return global; }
+    throw new Error('unable to locate global object');
+  })()
+  /*
+    The below code was added in case there is pollution in global namespace
+    windows object during transpilation with __esModule being set. Below code would support
+    global import in all bundle use cases
+  */
+  if (globalObj && globalObj.__esModule) {
+    globalObj.default = Object.assign({}, globalObj)
```
soumen-saniel, line 22: Can we find a way to not pollute the global `window` object and just have the `default` property set to the exported reference object.

soumen-saniel (changes requested):

soumen-saniel:
Please add more words to the PR description, explaining why `default` property is being set

lib/global/index.js
```diff
@@ -6,11 +6,36 @@
...
+          return target
+        }
+
+        return target[prop]
+      },
+
+      set: function (target, prop, value) {
+        target[prop] = value
+      },
+    };
```
soumen-saniel, line 34: Return a `Reflect` function from the get trap
And no need to trap the `set` method, let it be transparent

```javascript
get: function (target, prop, receiver) {
  if (prop === 'default') {
    return target
  }
  
  return Reflect.get(target, prop, receiver);
},
```

soumen-saniel (changes requested):

lib/global/index.js
Shubhakar-A: > Return a `Reflect` function from the get trap to preserve the correct scope. More details in https://javascript.info/proxy#reflect And no need to trap the `set` method, let it be transparent
> 
> ```js
> get: function (target, prop, receiver) {
>   if (prop === 'default') {
>     return target
>   }
>   
>   return Reflect.get(target, prop, receiver);
> },
> ```

This is done, 
Reflect.get(target, prop, receiver) - Throws a "illegal invocation javascript". This is due to reference change between proxy and actual windows when we try to invoce it. Replaced with Reflect.get(receiver, prop, target)

setter also same, used reflect here as well. Got  ""illegal invocation javascript"" error here as well without adding setter

soumen-saniel (approved):
Looks good

PR #134, opened 2023-08-11
fix: added boolean return for global set to fix Proxy set trapped error

Shubhakar-A:
# PR description

## JIRA tickets

[PD-215136](https://bazaarvoice.atlassian.net/browse/PD-215136)

## This PR is for

- Bug

## Requirements / issues

- BV loader throws ''set' on proxy: trap returned falsish for property javascript' error when latest global of bv-ui-core
- This is because, BV namespace has its own proxy and a setter which tries to set properties to window.BV
- global do not return any boolean in its setter method - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/Proxy/set#return_value

## Solutions

- Add boolean return for global's return method

## Unit tests

- N/A

## Version

- Patch - 2.8.2

[PD-215136]: https://bazaarvoice.atlassian.net/browse/PD-215136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/global/index.js
+++ b/lib/global/index.js
@@ -30,6 +30,7 @@ var getGlobal = function () {
 
       set: function (target, prop, value) {
         Reflect.set(target, prop, value)
+        return true
       },
     };
     const proxyGlobal = new Proxy(globalObj, override)
--- a/lib/global/index.js
+++ b/lib/global/index.js
@@ -28,8 +28,8 @@ var getGlobal = function () {
         return Reflect.get(receiver, prop, target)
       },
 
-      set: function (target, prop, value) {
-        Reflect.set(target, prop, value)
+      set: function (target, prop, value, receiver) {
+        Reflect.set(target, prop, value, receiver)
         return true
       },
     };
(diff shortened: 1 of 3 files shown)


soumen-saniel (changes requested):

lib/global/index.js
```diff
@@ -28,8 +28,9 @@ var getGlobal = function () {
         return Reflect.get(receiver, prop, target)
       },
 
-      set: function (target, prop, value) {
-        Reflect.set(target, prop, value)
+      set: function (target, prop, value, receiver) {
+        Reflect.set(target, prop, value, receiver)
+        return true
```
soumen-saniel, line 33: This should be just
```
set: function (target, prop, value, receiver) {
  return Reflect.set(target, prop, value, receiver)
},
```

soumen-saniel (changes requested):

soumen-saniel (approved):
Looks good

anishbvoice (approved):
Looks good

PR #137, opened 2024-03-25
PD-234259: TraegerService: Console error "Remove Child"

prashantbazaarvoice:
[PD-234259](https://bazaarvoice.atlassian.net/browse/PD-234259)

## This PR is for

- Bug

## Problem/Goal
The client is seeing an error in the console "Uncaught TypeError: Cannot read properties of null (reading 'removeChild')"

## Solution
As a fix now we are adding a condition if script is having parent node then we are allowing to remove the child by doing so it will not throw any error in the console.

**Before fix**

Screenshot 2024-03-25 at 11 26 15 AM

**After fix:**
 
Screenshot 2024-03-25 at 11 24 11 AM

Note: console error which is observed because clients are on lower version and we used redirection for testing so.

[PD-234259]: https://bazaarvoice.atlassian.net/browse/PD-234259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/loader/index.js
+++ b/lib/loader/index.js
@@ -152,8 +152,9 @@ module.exports = {
       done = true;
       clearTimeout(timeoutHandle);
       script.onload = script.onreadystatechange = script.onerror = null;
-      script.parentNode.removeChild(script);
-
+      if (script && script.parentNode) {
+        script.parentNode.removeChild(script);
+      }
       if (!err) {
         _loadedUrls[url] = true;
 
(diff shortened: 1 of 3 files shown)


vignesh-ph (approved):
Changes looks good

PR #146, opened 2025-01-07
allowing bvFetch to honor change the request headers while caching

VarshaAdigaBV:
# PR description

## JIRA tickets

[PD-260762](https://bazaarvoice.atlassian.net/browse/PD-260762)

## This PR is for

- Bug

## Requirements / issues

- The bvFetch module does not honor the change in headers while retrieving the cached response

## Solutions

- When storing the response in cache the request headers are not a part of the cache key. Making changes so that the cache key is a request so that when retrieving the cache from the cache storage the request headers are available.
- For BFD the response obj in the api response should be passed to the error handling function

[PD-260762]: https://bazaarvoice.atlassian.net/browse/PD-260762?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/bvFetch/index.js
+++ b/lib/bvFetch/index.js
@@ -10,41 +10,79 @@ module.exports = function BvFetch ({ shouldCache, cacheName, cacheLimit }) {
   this.cacheName = cacheName || 'bvCache';
   this.cacheLimit = cacheLimit * 1024 * 1024 || 10 * 1024 * 1024;
   this.fetchPromises = new Map();
-  this.cachedUrls = new Set();
+  this.cachedRequests = new Set();
 
   /**
-   * Generates a unique cache key for the given URL and options.
-   * @param {string} url - The URL of the API endpoint.
-   * @param {Object} options - Optional request options.
-   * @returns {string} The generated cache key.
+   * Checks if a request is present in a set of cached URLs.
+   *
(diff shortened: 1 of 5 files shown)


lib/bvFetch/index.js
```diff
@@ -10,41 +10,79 @@ module.exports = function BvFetch ({ shouldCache, cacheName, cacheLimit }) {
...
+  function isRequestInSet (cachedRequests, cacheKey) {
+    // Convert the Set to an array and check if any request matches
+    return [...cachedRequests].some((cachedRequest) => {
+      // Compare URLs
+      if (cachedRequest.url !== cacheKey.url) {
+        return false;
+      }
+  
+      // Compare headers
+      const cachedHeaders = [...cachedRequest.headers.entries()];
```
Shubhakar-A, line 33: I suggest add more null checks on object methods. -like .entries to avoid any page break errors. Or a try catch

Shubhakar-A (changes requested):

Shubhakar-A (approved):
Approved

visheshsanghani (approved):

